Issue #24, opened 2026-06-22
feat: introduce ProviderPermissions resource


OlegErshov:
## Description 
Add `ProviderPermissions` resource into security-operator and update AuthorizationModel generation flow to take care of relations defined in  `ProviderPermissions` resource and influence schema generation. 

## Changes
1. Add `ProviderPermissions` API
2. Adjust AuthorizationModel generation subroutine
3. Add `ProviderPermissions` ApiResourceSchema into platform-mesh-operator

PR #58, opened 2026-06-23
feat: add provider permissions resource

OlegErshov:

## Description 
This pr introduces ProviderPermissions resource based on this RFC https://github.com/platform-mesh/architecture/pull/42. 

TLDR
ProviderPermissions influences AuthorizationModel generation. It allows provider to override/add extra relations to the authorization schema. 

Changes:
1. Update AuthorizationModel template which is now populated with data from ProviderPermissions resource.
2. Add watcher on ProvierPermissions resource in ApiBinding controller to trigger AuthorizationModel generation when PP resource is changed/deleted/created

This pr is related to https://github.com/platform-mesh/platform-mesh/issues/24


apis/core/v1alpha1/providerpermissions_types.go
```diff
@@ -0,0 +1,142 @@
+package v1alpha1
+
+import (
+	"github.com/platform-mesh/subroutines/conditions"
```
xrstf: `go.platform-mesh.io/subroutines/conditions`

apis/providers/v1alpha1/providerpermissions_types.go
```diff
@@ -0,0 +1,142 @@
+package v1alpha1
```
xrstf, line 17: Please add the boilerplate from hack/boilerplate. The job to verify that is still TBD.

mjudeikis:
Im bit confused on the API we have here. We added providers api into pm-operator, is this related in any ways to that?

OlegErshov:
Hi, No these APIs covers different concepts. `Provider` and `ManagedProvider` API automate provider onborading and make it declarative. `ProviderPermissions` resource goal is to influence Authorization schema generation for the API provider exports. So provider should be onboarded already by `Provider`/`ManagedProvider` resources, and than when provider wants to change default AuthorizationModel, which has been generated for their resources, he can create `ProviderPermissions` resource to do that

mjudeikis:
> Hi, No these APIs covers different concepts. `Provider` and `ManagedProvider` API automate provider onborading and make it declarative. `ProviderPermissions` resource goal is to influence Authorization schema generation for the API provider exports. So provider should be onboarded already by `Provider`/`ManagedProvider` resources, and than when provider wants to change default AuthorizationModel, which has been generated for their resources, he can create `ProviderPermissions` resource to do that

Just of interest, can't we derive this data from APIExport itself? We have permissionClaims on them, and annotation telling its a "provider"?

OlegErshov:
I think no, because this resource should provide an ability for provider to override permissions in the way he wants, and not only override but add completely new permissions/relations even like this `define code_approver: admin or code_owner` all these data in permissions isn't related to the `PermissionsClaims`. If we use only ApiExport resource we need to set all of this in annotations which we don't want to do.

`ProviderPermissions` resource and `ApiExport` are connected by `ApiResourceSchema` resources which are listed in `ApiExport`. In the future we should have validation webhook which will do not allow updating permissions on the resources which are not introduced by related ApiExport

xrstf:
@OlegErshov the default branch of this repo has been reverted back to main, which in itself has been fast-forwarded to the monorepo branch. You should be able to simply switch this PR's target from monorepo to main.

operators/security-operator/internal/test/integration/suite_test.go
```diff
@@ -226,39 +267,70 @@ func (suite *IntegrationSuite) setupPlatformMesh(t *testing.T) {
 			return false
 		}
 		return len(endpointSlice.Status.APIExportEndpoints) > 0 && endpointSlice.Status.APIExportEndpoints[0].URL != ""
-	}, 10*time.Second, 200*time.Millisecond, "kcp should automatically create APIExportEndpointSlice with populated endpoints")
+	}, 10*time.Second, 200*time.Millisecond, "KCP should automatically create core.platform-mesh.io APIExportEndpointSlice with populated endpoints")
```
ntnn:
```suggestion
	}, 10*time.Second, 200*time.Millisecond, "kcp should automatically create core.platform-mesh.io APIExportEndpointSlice with populated endpoints")
```

ntnn (dismissed):

operators/security-operator/config/crd/bases/providers.platform-mesh.io_providerpermissions.yaml
```diff
@@ -0,0 +1,231 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.20.1
+  name: providerpermissions.providers.platform-mesh.io
```
mjudeikis, line 20: Any reason why we have {providers}permissions.{providers}.platfrom-mesh.io double "providers"? 

Usually, you either put it into a group or chose different group. We need to be more cautions about these names, as backend code we can change - apis, not so much.

mjudeikis (approved):
approving. 
I dont like this. Dont like this very much but I was told we have downstream waiting for this capability

PR #81, opened 2026-06-25
Wire in legacy jobs

mjudeikis:
## Summary by CodeRabbit

* **New Features**
  * Added a standardized container vulnerability scan workflow used across releases.
  * Added support for an additional legacy image publishing path for tag-based releases.

* **Bug Fixes**
  * Updated release workflows to use local scanning and publishing steps, improving consistency across projects.
  * Kept existing release and scan behavior intact while adjusting registry handling for older tag formats.

### DIFF
--- a/.github/workflows/backup-operator.yml
+++ b/.github/workflows/backup-operator.yml
@@ -201,7 +201,7 @@ jobs:
   trivy-image:
     if: github.ref == 'refs/heads/main'
     needs: [docker-build-latest]
-    uses: platform-mesh/.github/.github/workflows/job-trivy-image.yml@main
+    uses: ./.github/workflows/job-trivy-image.yml
     permissions:
       contents: read
       packages: read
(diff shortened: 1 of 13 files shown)


ntnn (approved):

.github/workflows/backup-operator.yml
```diff
@@ -332,6 +332,25 @@ jobs:
       # not the org root, to avoid cross-repo package permissions.
       ocmRegistryUrl: ghcr.io/platform-mesh/platform-mesh
 
+  image-ocm-legacy:
```
xmudrii, line 335: What is this pushing exactly? Do we need to add this job as a dependency to `trivy-sbom` (it requires the component to be pushed before it can be scanned)?

ntnn: It's the same image, it's just publishing to the old package

ntnn: Until we can update the ocm component constructors to reference the new locations

xmudrii: All good then, LGTM.

mjudeikis: This is so we can support monorepo + old OCM paths in helm-charts
And for now sboms are being published in "old" location. 

It a mess we have to have now :D

PR #87, opened 2026-06-25
Enable more Go linters globally

xrstf:
This extends the lists of linters that we apply for all components by those that were already pretty common (among our components), plus some that did not cause many changes but are IMHO very valuable (nilerr, nolintlint, reservedwords, ...).


golang-commons/errors/errors.go
```diff
@@ -121,8 +121,8 @@ func ancestorOfCause(ourStack []uintptr, causeStack errors.StackTrace) bool {
 	}
 
 	// We know the sizes are compatible, so compare program counters from back to front.
-	for idx := 0; idx < len(ourStack); idx++ {
-		if ourStack[len(ourStack)-1] != (uintptr)(causeStack[len(causeStack)-1]) {
+	for idx := range ourStack {
+		if ourStack[len(ourStack)-idx-1] != (uintptr)(causeStack[len(causeStack)-idx-1]) {
 			return false
 		}
 	}
```
xmudrii, line 128: The behavior of the new loop seems to differ, the old loop wasn't using `idx` at all, is this expected? Unless I'm missing something.

xrstf: Yes, I think the old code was broken, so I fixed it based on the comments in the function.

xmudrii (approved):

PR #118, opened 2026-06-26
Prevent go versioning mistakes

xrstf:
This PR adds a new script to ensure folks do not accidentally bump the `go` directive in the `go.mod` files when bumping Go for Platform Mesh. Too often have I seen this cause unnecessary headache for downstream consumers, especially of libraries.

This PR also changes the Go version used for testing/linting to be the "global" one, not the bare minimum. This is because within a Go workspace, the Go version dictated by the go.work has precedence and trying to lint on Go 1.25.0 when go.work is 1.25.7 would not work.

@xmudrii pointed out the caveat that this will prevent us from raising the minimum above what is required by the dependencies. If there ever is a Go feature that gets enabled by a specific Go version in the `go.mod`, and we would want to use it, we would have to adjust the script (or, what I think is more likely, bump dependencies once, since this so far has always bumped the required Go release ^^).


cmd/jl/go.mod
```diff
@@ -1,6 +1,6 @@
 module go.platform-mesh.io/jl
 
-go 1.25.0
+go 1.23
```
xmudrii, line 3: Is this downgrade expected?

xrstf: Yes, the script dictates it. I guesstimated 1.25 when integrating jl into this repo.

xmudrii (approved):

PR #144, opened 2026-06-29
begin reducing ctrlruntime dependencies in the APIs

xrstf:
The apis module should not have a dependency on controller-runtime, since CR is far too heavy and ties the apis to a specific Kube release, making re-use much harder than it has to be.

For this we need to undo some kubebuilder defaults, like relying on CR for scheme initialization and also later moving the webhook code out of the apis module.

This is the first step, harmonizing how the individual APIs are set up and not using CR for that anymore.

### DIFF
--- a/apis/migration/v1alpha1/groupversion_info.go
+++ b/apis/migration/v1alpha1/groupversion_info.go
@@ -20,17 +20,31 @@ limitations under the License.
 package v1alpha1
 
 import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	runtime "k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"sigs.k8s.io/controller-runtime/pkg/scheme"
+)
+
+const (
+	GroupName    = "migration.platform-mesh.io"
+	GroupVersion = "v1alpha1"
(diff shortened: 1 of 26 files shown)


apis/migration/v1alpha1/groupversion_info.go
```diff
@@ -20,17 +20,31 @@ limitations under the License.
...
+	runtime "k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"sigs.k8s.io/controller-runtime/pkg/scheme"
 )
 
-var (
-	// GroupVersion is group version used to register these objects
-	GroupVersion = schema.GroupVersion{Group: "migration.platform-mesh.io", Version: "v1alpha1"}
+const (
+	GroupName    = "migrastion.platform-mesh.io"
```
xmudrii: Type in the group name

```suggestion
	GroupName    = "migration.platform-mesh.io"
```

xmudrii (changes requested):

apis/migration/v1alpha1/groupversion_info.go
xrstf: Eagle Eye!

xmudrii (approved):

PR #147, opened 2026-06-29
Update to kcp 0.32.2, Kube 1.36

xrstf:
This bumps the monorepo to the latest kcp version and the Kube version that belongs to it.


xrstf:
This is failing because the e2e tests in PM first setup a WorkspaceType and then the other objects referenced by that WST. This is not allowed in kcp 0.32.1+ anymore (i.e. there is no more eventual consistency for WST's). I was told this concept is used in few places in PM, which makes the kcp 0.32 upgrade a lot harder.

ntnn (dismissed):
Neat.

But maybe we should wait until after tagging v0.4.0?

ntnn (dismissed):

ntnn (approved):

xmudrii (approved):

PR #149, opened 2026-06-30
fix(security-operator): enable org deletion via org workspace terminator

### LINKED ISSUE #43: bug: It's not possible to remove an organization
## Description
It's not possible to remove an organization because Terminator is enabled for an organization workspaces but there's no terminator handler. If it's desired to support organization cleanup, terminator either should be removed from org WorkspaceType or terminator handler should be implemented

As a side affect it's not possible to clean up core.platform-mesh.io ApiBinding in root:orgs workspace."

## Solution
1. Remove terminator from workspace type if it's not needed and all organization clean up logic is covered by Finalizers
2. Implement terminator controller for org workspace type

philtk79:
## Summary
- Register `OrgLogicalClusterTerminator` in the terminator deployment so org-type `LogicalCluster`s are reconciled during workspace deletion.
- Add `OrgTerminationSubroutine` to delete org-scoped `Store`, `IdentityProviderConfiguration`, and `WorkspaceAuthenticationConfiguration` in `root:orgs`, wait for finalizer-backed resources, then clear `root:security`.
- Treat a missing org OpenFGA store as a no-op in `AccountTuplesSubroutine.Terminate()` so orgs with child account workspaces can still be deleted cleanly.
Fixes #43.

## Problem
Org workspaces inherit the `root:security` terminator, but the security-operator terminator only handled account workspaces. Org deletion hung indefinitely because nothing removed the marker or cleaned up org init resources.
Deleting an org with sub-accounts could also stall when the org store was removed before child account workspaces finished tuple cleanup.


operators/security-operator/cmd/terminator.go
```diff
@@ -97,13 +98,29 @@ var terminatorCmd = &cobra.Command{
...
-			cmd.Context(),
 			fgaClient,
 			terminatorCfg.FGA.StoreIDCacheTTL,
+			cmd.Context(),
 			log,
 		)
 		kcpClientGetter := iclient.NewConfigSchemeKCPClientGetter(mgr.GetLocalManager().GetConfig(), mgr.GetLocalManager().GetScheme())
 
+		orgTerminatorCfg := terminatorCfg
+		orgTerminatorCfg.Initializer = config.InitializerConfig{}
```
OlegErshov: To do not make Initializer config empty and add an additional subroutine. We can Implement Terminate method in idp, store, workspace_authorization subroutines. Subroutine framework will call these methods when terminator name is set

philtk79: You are right, I implemented the terminate closer to the source

operators/security-operator/internal/subroutine/workspace_authorization.go
```diff
@@ -177,6 +178,29 @@ func (r *workspaceAuthSubroutine) reconcile(ctx context.Context, obj ctrlruntime
...
+	orgsClient, err := r.kcpClientGetter.NewClientForLogicalCluster(ctx, config.OrgsClusterPath)
+	if err != nil {
+		return subroutines.OK(), fmt.Errorf("getting orgs client: %w", err)
+	}
+
+	pending, err := deleteOrgResource(ctx, orgsClient, &kcptenancyv1alpha1.WorkspaceAuthenticationConfiguration{}, workspaceName)
+	if err != nil {
+		return subroutines.OK(), fmt.Errorf("deleting WorkspaceAuthenticationConfiguration %s: %w", workspaceName, err)
+	}
+	_ = pending // WAC has no finalizer; deletion is fire-and-forget during org termination.
```
OlegErshov: Let's better do this like `_, err := deleteOrgResource()` rather than do `_ = pernding` after

philtk79: Done

OlegErshov (approved):

PR #158, opened 2026-07-04
feat: add provider permissions role retriever

### LINKED ISSUE #213: Integrate ProviderPermissions into iam-service for roles lookup
## Description
Currently iam-service takes Roles for account and namespace resource from the hardcoded file. With ProviderPermissions we add an ability to introduce new Roles/relations and now iam-service needs to return roles based on the ProviderPermissions resource.

## Implementation
For that we need to implement a RoleRetriever interface based on the ProviderPermissions resource lookup

OlegErshov:

## Description
This PR adds roleRetriever implementation based on the ProviderPermissions resource. It uses multicluster runtime provider version from main branch due to a need of AggregateInformer.
The feature itself is hidden under the feature flag so a user can decide which retriever to use

This pr is related to https://github.com/platform-mesh/platform-mesh/issues/24


services/iam-service/go.mod
```diff
@@ -1,14 +1,14 @@
 module go.platform-mesh.io/iam-service
 
-go 1.25.7
+go 1.26.3
```
xrstf, line 3: We should wait for #147 so all components are on the same version, rather than diverging one component so drastically from the rest of the monorepo.

services/iam-service/pkg/roles/provider_permissions_cache.go
```diff
@@ -0,0 +1,138 @@
...
+
+func (c *ProviderPermissionsCache) GetRoles(groupResource string) []RoleDefinition {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+
+	roles, ok := c.roles[groupResource]
+	if !ok {
+		return nil
+	}
+	return roles
```
xrstf: A plain `return c.roles[groupResource]` should suffice, no? If you like explicit absence checking more, we can also keep it this way :)

services/iam-service/pkg/roles/provider_permissions_retriever.go
```diff
@@ -0,0 +1,54 @@
...
+	roles := r.cache.GetRoles(groupResource)
+	if roles == nil {
+		return []RoleDefinition{}, nil
+	}
+	return roles, nil
+}
+
+// buildGroupResource constructs the GroupResource key from group and kind.
+// Format: "{kind}.{group}" or just "{kind}" if group is empty.
+func buildGroupResource(group, kind string) string {
```
xrstf: I have a bellyache about this function turning "kinds" (singular) into "resources" (plural), but not actually applying any pluralization rules. This wouldn't be too bad on its own, if this was just internal cache keys, but the cache's API is `GetRoles(groupResource string)` -- so I would expect to be able to provide a key like "configmaps.core", but this function right here would instead construct "configmap.core".

I was wondering if `GetRoles()` should take a schema.GroupKind struct instead or if we should make this function here a public "key helper function". But even then, doing pluralization without a restmapper is brittle... maybe renaming the parameter to "groupKind" would help?

services/iam-service/pkg/roles/provider_permissions_cache.go
OlegErshov: Thanks! it's more clear to just return return c.roles[groupResource]

services/iam-service/pkg/roles/provider_permissions_retriever.go
OlegErshov: I think there's a little misunderstanding. This function only constructs groupResource string by combining kind and group of the resource in the format <kind>.<group> and if group is empty (like for namespaces), the function will return just kind.

We can use `schema.GroupKind` as a key for GetRoles() call as you've suggested, but we will need anyway somewhere do a parsing from/into <kind>.<group> format as this is the format from ProviderPermissions resource.

Also it's named groupResource just for consistency with file based role retriever

xrstf: I understand the need to create a stringified map key, my issue was the mix-up between "resources" and "kinds". When I see a `groupResource` parameter, I expect to be able to call it with, for example, `core.pods` (since `pods` is a _resource_ (i.e. lowercase and plural)), but the `buildGroupResource` function would turn it into `core.pod` (`pod` being a bad approximation by skipping the pluralization logic). So unexpectedly, `groupResource` is actually `groupLowerKind`.

So we should just make up our minds: Use resources or kinds for the map keys. But not have a function that takes a kind and then just pretends it's a resource name. :)

OlegErshov: thanks! I see it, I've updated the naming for consistency sake and used group-kind approach

services/iam-service/pkg/roles/provider_permissions_cache.go
```diff
@@ -0,0 +1,134 @@
...
+	"go.platform-mesh.io/golang-commons/logger"
+
+	toolscache "k8s.io/client-go/tools/cache"
+
+	"github.com/kcp-dev/multicluster-provider/pkg/provider"
+)
+
+type ProviderPermissionsCache struct {
+	mu sync.RWMutex
+	// roles maps GroupResource (e.g. "httpbin.orchestrate.platform-mesh.io") to role definitions
```
xrstf: Resources are plural, so let's turn the example into `httpbins.orchestrate.yadda.yadda`.

xrstf: Ah now I remember this PR 😅  If we settled on using GroupKind, then let's rename the parameters and comments in this file, too :)

OlegErshov: thanks! done

xrstf (approved):

Issue #128, opened 2026-06-20
feat(gateway): typed `resourcesByCategory(name)` query + subscription returning a union of category members


maximilianbraun:
## Motivation

Today the only cross-cutting query is `typeByCategory(name): [{group,version,kind,scope}]` — pure introspection, no instances. UI clients (e.g. the openmcp-project topology graph) need *instances of every GVK in a category* in one round-trip. The current options are both bad:

- **Client fan-out**: call `typeByCategory`, then issue N typed `list` queries in parallel and merge — 1+N round-trips, no subscription story, every consumer reinvents the merge.
- **Server-side BFF aggregator**: a separate REST service that re-implements what the gateway already knows about kinds/lists/watches.

## Proposal

For each category encountered during schema generation, emit a typed Union plus a query and subscription:

```graphql
union ManagedResource =
  CrossplaneXyzv1alpha1_VPC
  | CrossplaneXyzv1alpha1_Bucket
  | ... # one member per GVK in the category

type Query {
  resourcesByCategory(
    name: String!
    namespace: String
    labelSelector: String
    limit: Int
    continue: String
  ): [ManagedResource!]!
}

type Subscription {
  resourcesByCategory(
    name: String!
    namespace: String
    labelSelector: String
  ): ResourceEvent!
}

type ResourceEvent { type: String! object: ManagedResource! }
```

Clients use inline fragments + `__typename`, keeping codegen and cache normalization across consumers.

## Why a union (not a JSON scalar)

The gateway already builds a `*graphql.Object` per GVK in `gateway/schema/generator/generator.go` (`processResource`, ~line 269) and a `types.Registry` GVK→object map. A union just references those — no new type generation, no loss of type info, no client-side casting. `graphql-go` supports this via `graphql.NewUnion` + `ResolveTypeFn`.

## Implementation sketch

In `gateway/schema/extensions/`:
- Extend `CategoryManager` to also store the per-GVK `*graphql.Object` already created in the generator.
- Add `AddResourcesByCategoryQuery` next to `AddTypeByCategoryQuery` (`custom_queries.go:24`).
- Resolver = parallel fan-out wrapping the existing `resolver.ListItems` (`resolver.go:43`).
- Mirror for subscription via `subscription.go:75` `SubscribeItems`.
- List args mirror existing per-kind list resolver (`namespace`, `labelSelector`, `limit`, `continue`).

Estimated ~250 LoC + tests.

## Alternatives considered

[...]

PR #159, opened 2026-07-06
feat(gateway): Query and subscribe by category

brsmnv:
This PR adds two new operations for a custom query and a subscription by category, both named `resourcesByCategory`. 
A graphql union type `CategoryResource` is generated from the categorized resources. Resources not in a category are excluded.

During the PoC and test phase `resourcesByCategory` **requires a new feature flag** to be passed via CLI args: 
`--enable-resources-by-category`

Metrics added to the fan-out list and watch:
- Query:
  - `kubernetes_graphql_gateway_category_fanout_types_bucket`
  - `kubernetes_graphql_gateway_category_objects_returned_bucket`
- Subscribe:
  - `kubernetes_graphql_gateway_category_watches_active`
  - `kubernetes_graphql_gateway_watch_initial_items_bucket`

No new limits with defaults are added yet. We need feedback from a live system and the aforementioned metrics to make a better judgement call on that.

Related to #128

### DIFF
--- a/services/kubernetes-graphql-gateway/gateway/resolver/custom_queries.go
+++ b/services/kubernetes-graphql-gateway/gateway/resolver/custom_queries.go
@@ -24,7 +24,9 @@ import (
 	"golang.org/x/sync/errgroup"
 
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
 type TypeByCategory struct {
@@ -72,6 +74,10 @@ func (r *Service) ResourcesByCategory(m map[string][]TypeByCategory) graphql.Fie
 				scope := apiextensionsv1.ResourceScope(v.Scope)
 				items, err := r.ListItems(gvk, scope)(p)
 				if err != nil {
+					if apierrors.IsForbidden(err) {
+						log.FromContext(ctx).V(4).Info("skipping forbidden type in category", "category", name, "gvk", gvk, "err", err)
+						return nil
+					}
 					return fmt.Errorf("getting items for gvk %s: %w", gvk, err)
 				}
 
@@ -119,6 +125,8 @@ func (r *Service) SubscribeResourcesByCategory(m map[string][]TypeByCategory) gr
 			r.metrics.UpdateCategoryWatches(category, categoryTypesCount)
 		}
 
+		logger := log.FromContext(p.Context)
+
 		var wg sync.WaitGroup
 		for _, cType := range categoryTypes {
 			gvk := schema.GroupVersionKind{Group: cType.Group, Version: cType.Version, Kind: cType.Kind}
@@ -126,12 +134,12 @@ func (r *Service) SubscribeResourcesByCategory(m map[string][]TypeByCategory) gr
 
 			sub, err := r.SubscribeItems(gvk, scope)(p)
 			if err != nil {
-				// TODO: only ignore RBAC errors
+				logger.Error(err, "subscribing to type in category", "category", category, "gvk", gvk)
 				continue
 			}
 			srcChan, ok := sub.(chan any)
 			if !ok {
-				// TODO: error log
+				logger.Error(nil, "subscription source is not a channel", "type", fmt.Sprintf("%T", sub))
 				continue
 			}
 
@@ -141,6 +149,16 @@ func (r *Service) SubscribeResourcesByCategory(m map[string][]TypeByCategory) gr
 					case event, ok := <-srcChan:
 						if !ok {
 							return
+						}
+
+						if watchErr, isErr := event.(error); isErr {
+							if apierrors.IsForbidden(watchErr) {
+								logger.V(4).Info(
+									"skipping watch on forbidden type in category", "gvk", gvk, "err", watchErr)
+
+								// return without cancel is safe, srcChan is closed once an error is written
+								return
+							}
 						}
 						select {
 						case outCh <- event:
--- a/services/kubernetes-graphql-gateway/gateway/resolver/custom_queries.go
+++ b/services/kubernetes-graphql-gateway/gateway/resolver/custom_queries.go
@@ -17,7 +17,13 @@ limitations under the License.
 package resolver
 
 import (
+	"errors"
+	"fmt"
+
 	"github.com/graphql-go/graphql"
+
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
 type TypeByCategory struct {
@@ -35,5 +41,36 @@ func (r *Service) TypeByCategory(m map[string][]TypeByCategory) graphql.FieldRes
 		}
 
 		return m[name], nil
+	}
+}
+
+func (r *Service) ResourcesByCategory(m map[string][]TypeByCategory) graphql.FieldResolveFn {
+	return func(p graphql.ResolveParams) (any, error) {
+		name, err := GetArg[string](p.Args, NameArg, true)
+		if err != nil {
+			return nil, err
+		}
+
+		members := m[name]
+		result := make([]map[string]any, 0)
+
+		for _, v := range members {
+			gvk := schema.GroupVersionKind{Group: v.Group, Version: v.Version, Kind: v.Kind}
+			scope := apiextensionsv1.ResourceScope(v.Scope)
+			items, err := r.ListItems(gvk, scope)(p)
+			if err != nil {
+				return nil, fmt.Errorf("getting items: %w", err)
+			}
+
+			listresult, ok := items.(*ListResult)
+			if !ok {
+				return nil, errors.New("ListItems returned wrong type: was not *ListResult")
+			}
+
+			result = append(result, listresult.Items...)
+
+		}
+
+		return result, nil
 	}
 }
--- a/services/kubernetes-graphql-gateway/gateway/resolver/custom_queries.go
+++ b/services/kubernetes-graphql-gateway/gateway/resolver/custom_queries.go
@@ -87,9 +87,9 @@ func (r *Service) SubscribeResourcesByCategory(m map[string][]TypeByCategory) gr
 		categoryTypes := m[category]
 
 		var wg sync.WaitGroup
-		for _, cat := range categoryTypes {
-			gvk := schema.GroupVersionKind{Group: cat.Group, Version: cat.Version, Kind: cat.Kind}
-			scope := apiextensionsv1.ResourceScope(cat.Scope)
+		for _, cType := range categoryTypes {
+			gvk := schema.GroupVersionKind{Group: cType.Group, Version: cType.Version, Kind: cType.Kind}
+			scope := apiextensionsv1.ResourceScope(cType.Scope)
 
 			sub, err := r.SubscribeItems(gvk, scope)(p)
 			if err != nil {
--- a/services/kubernetes-graphql-gateway/gateway/resolver/custom_queries.go
+++ b/services/kubernetes-graphql-gateway/gateway/resolver/custom_queries.go
@@ -82,9 +82,8 @@ func (r *Service) SubscribeResourcesByCategory(m map[string][]TypeByCategory) gr
 			return nil, fmt.Errorf("no name arg in %v; %w", p.Args, err)
 		}
 
-		outCh := make(chan any, 100)
-
 		categoryTypes := m[category]
+		outCh := make(chan any, len(categoryTypes))
 
 		var wg sync.WaitGroup
 		for _, cType := range categoryTypes {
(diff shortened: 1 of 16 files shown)


services/kubernetes-graphql-gateway/gateway/resolver/custom_queries.go
```diff
@@ -37,3 +44,88 @@ func (r *Service) TypeByCategory(m map[string][]TypeByCategory) graphql.FieldRes
...
+				return nil, fmt.Errorf("getting items: %w", err)
+			}
+
+			listresult, ok := items.(*ListResult)
+			if !ok {
+				return nil, errors.New("ListItems returned wrong type: was not *ListResult")
+			}
+
+			result = append(result, listresult.Items...)
+		}
```
aaronschweig: Should we maybe use `errgroup.Group` to not have a sequential issue here? If we make the calls in parallel rather then sequential it probable scales a bit better for categories with a lot of resources

services/kubernetes-graphql-gateway/gateway/resolver/custom_queries.go
```diff
@@ -37,3 +44,88 @@ func (r *Service) TypeByCategory(m map[string][]TypeByCategory) graphql.FieldRes
...
+
+		var wg sync.WaitGroup
+		for _, cat := range categoryTypes {
+			gvk := schema.GroupVersionKind{Group: cat.Group, Version: cat.Version, Kind: cat.Kind}
+			scope := apiextensionsv1.ResourceScope(cat.Scope)
+
+			sub, err := r.SubscribeItems(gvk, scope)(p)
+			if err != nil {
+				// ignore (eg rbac?) or cancel everything?
+				continue
```
aaronschweig: we probably should not drop the error here

services/kubernetes-graphql-gateway/gateway/resolver/custom_queries.go
```diff
@@ -37,3 +44,88 @@ func (r *Service) TypeByCategory(m map[string][]TypeByCategory) graphql.FieldRes
...
+}
+
+func (r *Service) SubscribeResourcesByCategory(m map[string][]TypeByCategory) graphql.FieldResolveFn {
+	return func(p graphql.ResolveParams) (any, error) {
+		category, err := GetArg[string](p.Args, NameArg, true)
+		if err != nil {
+			return nil, fmt.Errorf("no name arg in %v; %w", p.Args, err)
+		}
+
+		outCh := make(chan any, 100)
```
aaronschweig: why do we limit the outgoing channel here? is the 100 a value with meaning?

services/kubernetes-graphql-gateway/gateway/schema/extensions/custom_queries.go
```diff
@@ -58,8 +69,115 @@ func (g *CustomQueryGenerator) AddTypeByCategoryQuery(rootQueryType *graphql.Obj
...
+		Types: unionTypes,
+		ResolveType: func(p graphql.ResolveTypeParams) *graphql.Object {
+			rawObj, ok := p.Value.(map[string]any)
+			if !ok {
+				return nil
+			}
+
+			apiVersion, ok := getField(rawObj, "apiVersion")
+			if !ok {
+				return nil
```
aaronschweig, line 180: is this intended that we swallow all the errors silently? maybe we should at least log anything here for proper detection if something is wrong

services/kubernetes-graphql-gateway/gateway/schema/extensions/custom_queries.go
```diff
@@ -58,8 +69,115 @@ func (g *CustomQueryGenerator) AddTypeByCategoryQuery(rootQueryType *graphql.Obj
...
+
+func getField(source map[string]any, field string) (string, bool) {
+	v, got := source[field]
+	if !got {
+		return "", got
+	}
+
+	s, ok := v.(string)
+	return s, ok
+}
```
aaronschweig: [nip] This seems like a helper method that somewhere somebody has probably written genericly already? I can imagine that either k/k has something upstream or we already have some helper like that somewhere to reuse.

services/kubernetes-graphql-gateway/gateway/schema/generator/generator.go
```diff
@@ -292,6 +300,7 @@ func (g *SchemaGenerator) processResource(
 		Name:   uniqueTypeName + "_Input",
 		Fields: inputFields,
 	})
+	g.typeRegistry.Register(uniqueTypeName, resourceType, inputType)
```
aaronschweig, line 317: qq: what does this line add?

aaronschweig (changes requested):
small first round of review

services/kubernetes-graphql-gateway/gateway/resolver/custom_queries.go
brsmnv: I didn't want to use an unbuffered channel on the fan-in path. Do we have any preferred values or configuration to use here?

services/kubernetes-graphql-gateway/gateway/resolver/custom_queries.go
brsmnv: This was a pure guess on my side and I need to test against a cluster. If the token has only partial permissions (eg get/list on some types in the category is not allowed), I would expect an error to surface here. I went with degraded response instead of a hard fail here, but even in that case an assertion / errors.Is would be necessary.  
What is the general preference wrt errors in subscribe? I didn't find anything which can signal this to the UI client without killing the subscription.

services/kubernetes-graphql-gateway/gateway/resolver/custom_queries.go
brsmnv: What limit should we set here? It will probably help, but the footgun grows larger and I'll need to replace the trivial append with something goroutine safe.

services/kubernetes-graphql-gateway/gateway/schema/generator/generator.go
brsmnv: The top-level `resourceType` was not in the `types.Registry` at all. Until now this was not needed apparently.
Right now im using it in the Union builder fn to look up the GraphQL object (schema?) by gvk:
- (1) get all GVKs by category
- (2) look up GraphQL schema by gvk: `gObj, _ := reg.Get(reg.GetUniqueTypeName(&gvk))`
- (3) drop in a map[gvk]: `typesByGVK[gvk] = gObj`

The lookup (2) came up empty as the root types were not in the registry at all. So building the union was not possible, at least not through the CategoryManager and the types Registry. Maybe there is another way to get everything needed which I missed?

The union is built at the end once all resources are processed. And we have nowhere to look up the GraphQL objects for the API types.

The obvious question is why it wasn't needed until now. The GraphQL object is being added to `ResourceContext` in `processResource` in this same loop where the just-constructed object is in scope:
```go
	g.typeRegistry.Register(uniqueTypeName, resourceType, inputType)
  
	rc := &fields.ResourceContext{
		ResourceType:   resourceType, // <- the graphql.Object
        // ...
  }

	g.queryGen.Generate(rc, queryVersionType)
	g.mutationGen.Generate(rc, mutationVersionType)
	g.subscriptionGen.Generate(rc, rootSubscription)
```

And later that same field is being accessed in `Generate()`, here e.g. in query.go[33]:
```go
	target.AddFieldConfig(rc.SingularName, &graphql.Field{
		Type:    graphql.NewNonNull(rc.ResourceType), // <- type wrapped here
		Args:    itemArgs,
		Resolve: g.resolver.GetItem(rc.GVK, rc.Scope),
	})
```

services/kubernetes-graphql-gateway/gateway/schema/extensions/custom_queries.go
brsmnv: Absent fields are not an error, but I replaced getField with `unstructured.NestedString`. It surfaces field-level issues (not a string) and this is an error worth flagging. ~~I'll see how I can get a logger in there.~~ 
Edit: error logged now

services/kubernetes-graphql-gateway/gateway/schema/extensions/custom_queries.go
brsmnv: Fix incoming, I found `unstructured.NestedString`

services/kubernetes-graphql-gateway/gateway/resolver/custom_queries.go
brsmnv: The only alternative I could come up with: channel buffer = len(categoryTypes) which equals the producer count. I think that this is a good middleground between no buffer (writes constantly blocked on the fan-in) and huge buffer **per-subscription-query** (memory footprint goes up).

brsmnv:
This will need a feature flag at the very least, as the current limits are ineffective. On the HTTP side we have a fan-in taking all the subscriptions for a category, resulting in a single connection against the limit.

brsmnv:
Some metrics examples from my tests. Request duration is already covered by generic HTTP metrics. I think that these four are enough for the PoC.

**List**
```
# HELP kubernetes_graphql_gateway_category_fanout_types Number of resource types a category query expanded out to, by category.
# TYPE kubernetes_graphql_gateway_category_fanout_types histogram
kubernetes_graphql_gateway_category_fanout_types_bucket{category="cert-manager",le="1"} 0
kubernetes_graphql_gateway_category_fanout_types_bucket{category="cert-manager",le="2"} 0
kubernetes_graphql_gateway_category_fanout_types_bucket{category="cert-manager",le="4"} 0
kubernetes_graphql_gateway_category_fanout_types_bucket{category="cert-manager",le="8"} 1
(...)
kubernetes_graphql_gateway_category_fanout_types_sum{category="cert-manager"} 6
kubernetes_graphql_gateway_category_fanout_types_count{category="cert-manager"} 1

----------------------

# HELP kubernetes_graphql_gateway_category_objects_returned Total objects returned across all types in a category query, by category.
# TYPE kubernetes_graphql_gateway_category_objects_returned histogram
kubernetes_graphql_gateway_category_objects_returned_bucket{category="cert-manager",le="1"} 0
kubernetes_graphql_gateway_category_objects_returned_bucket{category="cert-manager",le="4"} 0
kubernetes_graphql_gateway_category_objects_returned_bucket{category="cert-manager",le="16"} 0
kubernetes_graphql_gateway_category_objects_returned_bucket{category="cert-manager",le="64"} 0
kubernetes_graphql_gateway_category_objects_returned_bucket{category="cert-manager",le="256"} 1
(...)
kubernetes_graphql_gateway_category_objects_returned_sum{category="cert-manager"} 68
kubernetes_graphql_gateway_category_objects_returned_count{category="cert-manager"} 1
```

**Subscribe**
```
# HELP kubernetes_graphql_gateway_category_watches_active Current number of open resource watches held by category subscriptions.
# TYPE kubernetes_graphql_gateway_category_watches_active gauge
kubernetes_graphql_gateway_category_watches_active{category="cert-manager"} 6

----------------------

# HELP kubernetes_graphql_gateway_watch_initial_items Number of objects returned during a subscription's list phase.
# TYPE kubernetes_graphql_gateway_watch_initial_items histogram
kubernetes_graphql_gateway_watch_initial_items_bucket{kind="Certificate",le="1"} 0
kubernetes_graphql_gateway_watch_initial_items_bucket{kind="Certificate",le="4"} 0
kubernetes_graphql_gateway_watch_initial_items_bucket{kind="Certificate",le="16"} 0
kubernetes_graphql_gateway_watch_initial_items_bucket{kind="Certificate",le="64"} 1
kubernetes_graphql_gateway_watch_initial_items_bucket{kind="Certificate",le="256"} 1
(...)
kubernetes_graphql_gateway_watch_initial_items_sum{kind="Certificate"} 28
kubernetes_graphql_gateway_watch_initial_items_count{kind="Certificate"} 1
kubernetes_graphql_gateway_watch_initial_items_bucket{kind="CertificateRequest",le="1"} 0
kubernetes_graphql_gateway_watch_initial_items_bucket{kind="CertificateRequest",le="4"} 0
kubernetes_graphql_gateway_watch_initial_items_bucket{kind="CertificateRequest",le="16"} 0
kubernetes_graphql_gateway_watch_initial_items_bucket{kind="CertificateRequest",le="64"} 1
(...)
kubernetes_graphql_gateway_watch_initial_items_sum{kind="CertificateRequest"} 28
kubernetes_graphql_gateway_watch_initial_items_count{kind="CertificateRequest"} 1
```

services/kubernetes-graphql-gateway/gateway/resolver/custom_queries_test.go
```diff
@@ -0,0 +1,399 @@
...
+			case error:
+				err = v
+			case SubscriptionEnvelope:
+				obj, _ := v.Object.(map[string]any)
+				name, _, _ := unstructured.NestedString(obj, "metadata", "name")
+				names = append(names, name)
+			default:
+				t.Fatalf("unexpected event type %T", event)
+			}
+		case <-time.After(200 * time.Millisecond):
```
aaronschweig: [nip] this seems like smth that can cause flaky tests in the future

aaronschweig (dismissed):
One minor comment. But overall looks good and can serve as a good PoC implementation to gather some data about the impact that the fanout has

aaronschweig (approved):

PR #166, opened 2026-07-08
fix(search-service): treat empty query as match-all instead of rejecting with 400

lpgarzonr:
## Summary

- An empty or whitespace-only `q` parameter now maps to a wildcard (`*`) query instead of returning a 400 error
- This aligns with standard OpenSearch/Elasticsearch conventions where an empty search returns all results
- Updated `TestSearchRejectsMissingQuery` → `TestSearchEmptyQueryMatchesAll` to verify the new behavior

## Motivation

The frontend sends `q=` on initial page load (before the user types anything). The previous behavior of returning a 400 forced callers to special-case empty queries. The fix makes the service consistent with OpenSearch's own behavior where an empty query string is equivalent to `*` (match-all).

## Test plan

- [x] `TestSearchEmptyQueryMatchesAll` verifies whitespace-only query returns results
- [x] All existing tests pass: `go test ./internal/service/search/...`

### DIFF
--- a/services/search-service/internal/service/search/service_test.go
+++ b/services/search-service/internal/service/search/service_test.go
@@ -127,11 +127,21 @@ func TestSearchInvalidCursor(t *testing.T) {
 	}
 }
 
-func TestSearchRejectsMissingQuery(t *testing.T) {
-	svc := NewService(fakeResolver{}, &fakeSearcher{}, &fakeAuthorizer{}, nil, ServiceConfig{})
-	_, err := svc.Search(context.Background(), SearchRequest{Organization: "acme", User: "alice@example.com", Query: "  "})
-	if !errors.Is(err, ErrInvalidRequest) {
-		t.Fatalf("expected ErrInvalidRequest, got %v", err)
+func TestSearchEmptyQueryMatchesAll(t *testing.T) {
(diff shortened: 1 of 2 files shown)


services/search-service/internal/service/search/service_test.go
```diff
@@ -127,11 +127,21 @@ func TestSearchInvalidCursor(t *testing.T) {
...
+func TestSearchEmptyQueryMatchesAll(t *testing.T) {
+	hit := OpenSearchHit{ID: "1", Score: 1, Sort: []any{1.0, "1"}, Source: map[string]any{"id": "1"}}
+	svc := NewService(
+		fakeResolver{index: SearchIndexRef{IndexName: "idx-acme", Resource: "accounts"}},
+		&fakeSearcher{pages: []OpenSearchPage{{Hits: []OpenSearchHit{hit}}}},
+		&fakeAuthorizer{results: []AuthorizationResult{{Allowed: []bool{true}, Calls: 1}}},
+		nil,
+		ServiceConfig{},
+	)
+	resp, err := svc.Search(context.Background(), SearchRequest{Organization: "acme", User: "alice@example.com", Query: "  "})
```
xrstf: Could use this opportunity to switch to `t.Context()`.

lpgarzonr: Good call! i updated in the latest commit

FWuermse (approved):

xrstf (approved):

PR #168, opened 2026-07-09
Add Tilt local dev environment (infra, kcp, account-operator)

mjudeikis:
Local-dev Tilt setup per : cert-manager, envoy gateway, etcd, dex, kcp-operator + static kcp (RootShard/FrontProxy), Flux + OCM delivery engines, and account-operator as a hot-reloadable L1 component. Profiles map to RFC 008 layers (infra/core/auth/full).

Further split is not possible for too much interconnections between platfrom-mesh-operator, bootstrap processes(es) and individual components. To move this further we need to make every component self sustainable, and independent as much as they can be. First thing first - split core.platfrom-mesh.io so it does not contain multiple component apis.

### DIFF
--- a/contrib/tilt/Tiltfile
+++ b/contrib/tilt/Tiltfile
@@ -1,0 +1,243 @@
+# Copyright 2026 The Platform Mesh Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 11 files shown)


contrib/tilt/Tiltfile
```diff
@@ -0,0 +1,243 @@
...
+
+# Profiles map onto RFC 008 shapes (see ADR). Skeleton implements `infra`.
+config.define_string('profile', args=False, usage='infra|core|auth|full (default: infra)')
+config.define_string_list('components', args=True, usage='monorepo components to hot-reload')
+cfg = config.parse()
+PROFILE = cfg.get('profile', 'infra')
+COMPONENTS = cfg.get('components', [])
+
+# Base domain + namespace shared by the whole environment.
+BASE_DOMAIN = 'kcp.localhost'
```
ntnn: Maybe pm.localhost, place kcp components under kcp.BASE_DOMAIN and other things like Dex under dex.BASE_DOMAIN, so it's kcp.pm.localhost, dex.pm.localhost, ...?

ntnn (approved):

PR #172, opened 2026-07-09
feat: allow providers to configure the search for their schemas

FWuermse:
Implements support for the search configuration as per [ADR-007](https://github.com/platform-mesh/architecture/blob/main/adr/007-search-config-resource.md) and ticket https://github.com/platform-mesh/backlog/issues/279.

Providers can configure which and how fields of their own APIResourceSchemas should be indexed. Default indexes all lexically.

On-behalf-of: https://github.com/SAP [florian.wuermseer@sap.com](mailto:florian.wuermseer@sap.com)

Also refactors responsibility of SearchIndex resource:
1. Binding Reconciler is now responsible for flattening hierarchy into SearchIndex
2. SearchIndex Reconciler replicates the field into Open Search
3. Indexing merely uses what Open Search configured for the given field

Fields from an APIResourceSchema can be:
- Made a filter (exact match)
- Made searchable via vector search (semantic search)
- Excluded from the Index completely

Example:

- Provider exports APIResourceSchema `apidefinitions`
- Provider has SearchConfig resource also named `apidefinitions`
- SearchConfig assigns field `spec.description` as semantic
- Open Search let's us perform vector search on that field
image


operators/search-operator/internal/subroutine/apibinding_watcher.go
```diff
@@ -179,25 +190,112 @@ func (s *apiBindingWatcherSubroutine) resolveDefaultFields(ctx context.Context,
...
+	return result
+}
+
+func toSet(slice []string) map[string]bool {
+	m := make(map[string]bool, len(slice))
+	for _, s := range slice {
+		m[s] = true
+	}
+	return m
+}
```
xrstf: apimachinery's `set.Set` type could do this for us already.

operators/search-operator/internal/subroutine/apibinding_watcher.go
```diff
@@ -179,25 +190,112 @@ func (s *apiBindingWatcherSubroutine) resolveDefaultFields(ctx context.Context,
...
+
+func contains(slice []string, s string) bool {
+	for _, v := range slice {
+		if v == s {
+			return true
+		}
 	}
-	sort.Strings(fields)
-	return fields, nil
+	return false
```
xrstf: Is this just `slices.Contains(slice, s)`?

operators/search-operator/internal/subroutine/apibinding_watcher.go
```diff
@@ -139,65 +149,184 @@ func (s *apiBindingWatcherSubroutine) Finalize(_ context.Context, _ runtimeobjec
...
+	}
+
+	var matched *kcpapisv1alpha1.APIResourceSchema
+	for i := range schemaList.Items {
+		s := &schemaList.Items[i]
+		if s.Spec.Names.Plural == pc.Resource {
+			matched = s
+			break
 		}
+	}
```
xrstf, line 193: Over time, resources get updated and so more and more APIResourceSchemas for the same resource will exist, with the APIExport pointing to the "current" version that should be used.

How will this code cope with that? If someone has

* v1.things.example.corp
* v2.things.example.corp
* v3.things.example.corp
* v4.things.example.corp

.. then this controller would always only work with v1, no?

xrstf: (v1..v4 not being the Kubernetes API versions, but the APIResourceSchema versions, meaning all four of them could define "v1alpha1" and "v1beta1", or maybe v4 contains additionally v1beta2)

operators/search-operator/internal/subroutine/apibinding_watcher.go
```diff
@@ -139,65 +149,184 @@ func (s *apiBindingWatcherSubroutine) Finalize(_ context.Context, _ runtimeobjec
...
-		for _, version := range schema.Spec.Versions {
-			if !version.Served {
-				continue
-			}
-			props, err := version.GetSchema()
-			if err != nil {
-				return nil, fmt.Errorf("parse schema for %q version %q: %w", br.Schema.Name, version.Name, err)
+	defaultFields := make(map[string]struct{})
+	semanticFields := make(map[string]struct{})
+	exactFields := make(map[string]struct{})
```
xrstf, line 201: Can't we use apimachinery's Set type? Then `sliceFromSet` can go away and be replaced with `sets.List(defaultFields)`, for example. This also has the advantage of being sorted, i.e. consistent.

operators/search-operator/internal/subroutine/apibinding_watcher.go
```diff
@@ -228,18 +359,24 @@ func (s *apiBindingWatcherSubroutine) ensureSearchIndex(
...
 			Msg("created SearchIndex")
 
 	case err != nil:
 		return fmt.Errorf("get SearchIndex %q: %w", searchIndexName, err)
 
 	default:
-		if stringSlicesEqual(existing.Spec.DefaultFields, defaultFields) {
+		if stringSlicesEqual(existing.Spec.DefaultFields, fields.defaultFields) &&
+			stringSlicesEqual(existing.Spec.SemanticFields, fields.semanticFields) &&
+			stringSlicesEqual(existing.Spec.FilterableFields, fields.filterableFields) {
```
xrstf, line 373: While we're at it, let's replace this with `slices.Equal()` and get rid of stringSlicesEqual alltogether.

apis/search/v1alpha1/searchconfig_types.go
```diff
@@ -0,0 +1,98 @@
...
+
+	// SemanticFields lists field paths to be indexed for vector/semantic search.
+	// These are typically human-readable text fields where meaning matters.
+	// +optional
+	SemanticFields []string `json:"semanticFields,omitempty"`
+
+	// ExactFields lists field paths to be indexed as keyword for exact-match filtering/faceting.
+	// These are typically identifiers, enums, or structured values.
+	// +optional
+	ExactFields []string `json:"exactFields,omitempty"`
```
xrstf, line 49: Sometimes, late at night, I see and write code like this and then wonder: Wouldn't it be nice if I could just use something like jq to transform one JSON object to another?

So I wonder if using something like https://jsonata.org/ would be overkill but also super fancy. No more us implementing basic JSONPaths, instead we just take a script like

```
{
   "exact": {
       "customerName": $.spec.customer.name,
       "customerPhone": $.spec.contacts[0].phone
   },
   "semantic": {
       "address": $.spec.address
   }
}
```

and thereby give providers much more freedom in slicing and dicing their objects to be efficiently indexable. Similar to CEL, just be a bit different in concept.

---

Then I wake up and am reminded that I pitch such ideas everywhere and nobody wants it :D

operators/search-operator/internal/subroutine/apibinding_watcher.go
FWuermse: That's a good point. Do you know how we can get the version from the Permission Claim? The kcp type only gives us the GR for the Resources in the Permission Claims but no version.

PR #194, opened 2026-07-15
fix: clustercache engage problems

### LINKED ISSUE #190: AuthZ Webhook: One APIBinding with unaccepted logicalclusters permission claim wedges the whole cluster cache
## Issue description

If a workspace's APIBinding to core.platform-mesh.io hasn't accepted the logicalclusters.core.kcp.io permission claim, the webhook can't read that workspace's LogicalCluster. Instead of skipping it, Engage retries forever — and because it blocks, every other workspace on the same shard stops getting cached too. Those requests then fall through to NoOpinion, so a lot of authz requests silently stop being served.

It doesn't recover on its own, even after the binding is fixed. Only a webhook restart clears it.

## Root cause

clusterCache.Engage  (pkg/clustercache/clustercache.go#L78) polls for the LogicalCluster with wait.PollUntilContextCancel and swallows every error (return false, nil), so it never gives up and never returns.

But multicluster-runtime's Engage contract explicitly requires the opposite:

Implementers should return an error if they cannot start operations for the given Cluster, and should ensure this operation is re-entrant and non-blocking.

Engage is called synchronously on the provider's single per-shard informer goroutine, so one blocked Engage head-of-line-blocks discovery for the entire shard. Result: cache never populates → cache miss in contextual.Handle → retries exhaust → NoOpinion.
.
## Suggested fix

Make Engage non-blocking and fail-fast, per the contract:

Bound the wait (PollUntilContextTimeout) and return err instead of return false, nil, so a single bad workspace is dropped without stalling the others; or
Move the LogicalCluster resolution off the Engage call path (goroutine or a reactive watch/reconciler that fills the cache).
Optionally pre-filter workspaces whose logicalclusters claim isn't accepted so they never reach Engage.

## Impact

Any single misconfigured/incomplete APIBinding degrades authorization for all workspaces sharing that kcp shard. Failure is silent (only a V(5) debug log) and requires a restart to clear.

aaronschweig:
Probably worth investigating if it makes sense to introduce a `max-concurrent-reconciles` equivalent to not nuke the operator on restart

Closes #190

### DIFF
--- a/services/rebac-authz-webhook/pkg/clustercache/clustercache.go
+++ b/services/rebac-authz-webhook/pkg/clustercache/clustercache.go
@@ -51,22 +51,52 @@ type Provider interface {
 	Get(clusterName multicluster.ClusterName) (ClusterInfo, bool)
 }
 
-type clusterCache struct {
-	lock  sync.RWMutex
-	cache map[multicluster.ClusterName]ClusterInfo
-	mgr   mcmanager.Manager
+var defaultBackoff = wait.Backoff{
+	Duration: time.Second,
+	Factor:   2.0,
+	Jitter:   0.1,
+	Steps:    9,
+	Cap:      5 * time.Minute,
 }
 
-func New(mgr mcmanager.Manager) (*clusterCache, error) {
-	return &clusterCache{
-		cache: make(map[multicluster.ClusterName]ClusterInfo),
-		mgr:   mgr,
-	}, nil
+type resolver struct {
+	cl     cluster.Cluster
+	cancel context.CancelFunc
+}
+
+type clusterCache struct {
+	cacheLock sync.RWMutex
+	cache     map[multicluster.ClusterName]ClusterInfo
+
+	resolversLock sync.Mutex
+	resolvers     map[multicluster.ClusterName]*resolver
+
+	mgr     mcmanager.Manager
+	backoff wait.Backoff
+}
+
+type Option func(*clusterCache)
+
+func WithBackoff(b wait.Backoff) Option {
+	return func(c *clusterCache) { c.backoff = b }
+}
+
+func New(mgr mcmanager.Manager, opts ...Option) (*clusterCache, error) {
+	c := &clusterCache{
+		cache:     make(map[multicluster.ClusterName]ClusterInfo),
+		resolvers: make(map[multicluster.ClusterName]*resolver),
+		mgr:       mgr,
+		backoff:   defaultBackoff,
+	}
+	for _, opt := range opts {
+		opt(c)
+	}
+	return c, nil
 }
 
 func (c *clusterCache) Get(clusterName multicluster.ClusterName) (ClusterInfo, bool) {
-	c.lock.RLock()
-	defer c.lock.RUnlock()
+	c.cacheLock.RLock()
+	defer c.cacheLock.RUnlock()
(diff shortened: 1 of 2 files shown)


services/rebac-authz-webhook/pkg/clustercache/clustercache.go
```diff
@@ -51,46 +51,123 @@ type Provider interface {
...
-		lc.SetGroupVersionKind(schema.GroupVersionKind{
-			Group:   "core.kcp.io",
-			Version: "v1alpha1",
-			Kind:    "LogicalCluster",
-		})
-		if err := cl.GetClient().Get(ctx, types.NamespacedName{Name: "cluster"}, &lc); err != nil {
-			klog.V(5).ErrorS(err, "Failed to get LogicalCluster, will retry", "clusterName", name)
-			return false, nil
+	c.resolversLock.Lock()
+	defer c.resolversLock.Unlock()
```
xrstf, line 108: Let's leave a comment like

```
// Handle cluster discovery asynchronously in order to not block Engage() calls on
// clusters with (hopefully just temporarily) missing permission claims.
```

services/rebac-authz-webhook/pkg/clustercache/clustercache.go
```diff
@@ -51,46 +51,123 @@ type Provider interface {
...
+	go func() {
+		defer func() {
+			c.resolversLock.Lock()
+			if cur, ok := c.resolvers[name]; ok && cur == r {
+				delete(c.resolvers, name)
+			}
+			c.resolversLock.Unlock()
+		}()
+		c.resolve(resolveCtx, name, cl)
+	}()
```
xrstf: Personally I find this defer usage a bit confusing and would suggest

```go
c.resolve(resolveCtx, name, cl)

c.resolversLock.Lock()
defer c.resolversLock.Unlock()

// cleanup if we're still active and have not been replaced by another Engage() call
if cur, ok := c.resolvers[name]; ok && cur == r {
	delete(c.resolvers, name)
}
```

Super technically the "ok" check is redundant since the empty value for a resolver should never be the same as a real resolver, so this can be even shortened to

```go
c.resolve(resolveCtx, name, cl)

c.resolversLock.Lock()
defer c.resolversLock.Unlock()

// cleanup if we're still active and have not been replaced by another Engage() call
if c.resolvers[name] == r {
	delete(c.resolvers, name)
}
```

services/rebac-authz-webhook/pkg/clustercache/clustercache.go
aaronschweig: done

services/rebac-authz-webhook/pkg/clustercache/clustercache.go
aaronschweig: makes sense, I adapted your suggestions - makes it way cleaner

BergCyrill (approved):
(On request by Aaron):
This fixes the issue for me for now.
For future improvements:
- use standing constants like the kcp annotation for the lc path instead of providing it inline
- same for own constants like GVK of platform-mesh resources
- On cold-start / during cache-warmup there may be alot (= # workspaces) simultaneous calls against kcp - memory footprint should be small so I would not expect issues there.

OlegErshov (approved):

PR #218, opened 2026-07-28
feat(kubernetes-graphql-gateway): make QPS and Burst configurable via arguments

ogligorov:
The kubernetes-graphql-gateway uses the client-go REST client to talk to Kubernetes API servers. QPS and Burst are left at client-go defaults (5 QPS / 10 burst), with no way to tune them.

The PR adds two CLI flags to expose these settings:

--kubernetes-qps - max requests per second (default 0, which preserves the client-go default of 5)
--kubernetes-burst - max burst size (default 0, which preserves the client-go default of 10)
The values are threaded from ExtraOptions → gateway.Config → endpoint.New → cluster.New, where they are applied to the rest.Config before any clients are built.

Setting either flag to 0 keeps the existing client-go default behaviour.

### DIFF
--- a/services/kubernetes-graphql-gateway/README.md
+++ b/services/kubernetes-graphql-gateway/README.md
@@ -201,8 +201,8 @@ Optionally set `ca.secretRef` for custom CA certificates.
 | `--max-query-batch-size` | `10` | Max queries per batch request |
 | `--read-header-timeout` | `32s` | Max duration for reading request headers |
 | `--idle-timeout` | `90s` | Max idle duration for keep-alive connections |
-| `--kubernetes-qps` | `0` (client-go default: 5) | Max requests per second to the Kubernetes API server |
-| `--kubernetes-burst` | `0` (client-go default: 10) | Max burst size for requests to the Kubernetes API server |
+| `--kubernetes-qps` | `0` (uses client-go default: 5) | Max requests per second to the Kubernetes API server |
+| `--kubernetes-burst` | `0` (uses client-go default: 10) | Max burst size for requests to the Kubernetes API server |
 
 Set any limit flag to `0` to disable that limit.
 
--- a/services/kubernetes-graphql-gateway/README.md
+++ b/services/kubernetes-graphql-gateway/README.md
@@ -201,6 +201,8 @@ Optionally set `ca.secretRef` for custom CA certificates.
 | `--max-query-batch-size` | `10` | Max queries per batch request |
 | `--read-header-timeout` | `32s` | Max duration for reading request headers |
 | `--idle-timeout` | `90s` | Max idle duration for keep-alive connections |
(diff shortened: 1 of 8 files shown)


xrstf:
LGTM, but please signoff your commit to make the DCO check pass.

aaronschweig (changes requested):
Thanks for the PR. A few blocking issues before this can merge:

  **1. The CLI flags are never registered.**
  `KubernetesQPS`/`KubernetesBurst` are added to `ExtraOptions` but there are no `fs.Float32Var`/`fs.IntVar` calls in `AddFlags` (`options.go`). As written, `--kubernetes-qps` and `--kubernetes-burst` don't exist — passing them fails with "unknown flag" and the fields stay at zero. Add:

  ```go
  fs.Float32Var(&options.KubernetesQPS, "kubernetes-qps", options.KubernetesQPS, "max requests per second to the Kubernetes API server (0 uses client-go default)")
  fs.IntVar(&options.KubernetesBurst, "kubernetes-burst", options.KubernetesBurst, "max burst size for requests to the Kubernetes API server (0 uses client-go default)")
  ```

  **2. `QPS = 0` disables rate limiting, it does not preserve the default.**
  The description/README claim `0` preserves the client-go default (5 QPS / 10 burst), but we explicitly assign `cluster.restCfg.QPS = 0`. Only assign when non-zero so client-go falls back to its default limiter:

  ```go
  if kubernetesQPS > 0 {
      cluster.restCfg.QPS = kubernetesQPS
  }
  if kubernetesBurst > 0 {
      cluster.restCfg.Burst = kubernetesBurst
  }
  ```

  Note the trade-off: with this guard, `0` means "use the client-go default (5/10)" — there is then **no** value that fully disables rate limiting, since `RESTClientFor` treats an unset/zero `QPS` as "apply default limiter", not "unlimited". If we want an explicit "disable" option, we'd need a dedicated sentinel that sets `RateLimiter = flowcontrol.NewFakeAlwaysRateLimiter()` rather than relying on `0`. Please decide the intended semantics and align the flag help/README accordingly.

  **3. Missing validation.**
  Every other numeric option has a `must not be negative` guard in `Validate()`. Add the same for `KubernetesQPS` and `KubernetesBurst`.

  **4. No tests.**
  `cluster.New` changed signature and gained rate-limit logic. Add a unit test asserting non-zero values are applied to `restCfg`/`adminCfg` and zero values leave client-go defaults intact.

  **5. Prefer an options struct over growing positional args.**
  This PR adds two more positional parameters to `cluster.New` (and threads similar additions through `endpoint.New`). Rather than extending these signatures indefinitely, consider introducing an options/config struct (e.g. `cluster.Options`) so future tuning knobs can be added without churning every call site and to keep the signatures readable

aaronschweig (approved):

PR #219, opened 2026-07-28
feat: add new authorization  endpoint

OlegErshov:

## Description
This pr adds an authorization batch endpoint for authzorization webhook. This endpoint should serve for the frontend sake to allow batch checks on many resources on the page. 

## Details
1. for batch endpoint a new http server has been added to the webhook, because server from manager's webhook works as an https server. To be aligned with iam-service and graphql gateway we need http server
2. if a check item is corrupt, as a result `allowed:false` will be return

## Request
Webhook expect the request like this 

```
 curl -X https://portal.localhost:8443/authz/batch \
      -H "Content-Type: application/json" \
      -d '{
      "items": [
        {
          "id": "check-1",
          "user": "john.doe@example.com",
          "clusterPath": "root:orgs:test:my-cluster",
          "resourceAttributes": {
            "verb": "get",
            "group": "apps",
            "resource": "deployments",
            "namespace": "default",
            "name": "my-deployment"
          }
        },
        {
          "id": "check-2",
          "user": "john.doe@example.com",
          "clusterPath": "root:orgs:test:my-cluster",
          "resourceAttributes": {
            "verb": "list",
            "group": "",
            "resource": "pods",
            "namespace": "kube-system"
          }
        },
        {
          "id": "check-3",
          "user": "jane.smith@example.com",
          "clusterPath": "root:orgs:test:my-cluster",
          "resourceAttributes": {
            "verb": "create",
          "user": "john.doe@example.com",
          "clusterPath": "root:orgs:test:my-cluster",
          "resourceAttributes": {
            "verb": "list",
            "group": "",
            "resource": "pods",
            "namespace": "kube-system"
          }
        },
        }
      ]
    }'
```

## Response 
```json
  {
    "results": [
      {"id": "check-1", "allowed": true},
      {"id": "check-2", "allowed": false},
      {"id": "check-3", "allowed": true}
    ]
  }
  ```

### CODE REVIEW

File: services/rebac-authz-webhook/cmd/serve.go [hunk 1]
```diff
@@ -154,3 +169,57 @@ func NewServeCmd() *cobra.Command {
...
+		idleTimeout:     cfg.BatchAuthzIdleTimeout,
+		shutdownTimeout: cfg.BatchAuthzShutdownTimeout,
+	}
+}
+
+func (s *authzServer) Start(ctx context.Context) error {
+	mux := http.NewServeMux()
+	mux.Handle("/batch-authz", s.handler)
+
+	server := &http.Server{
```
aaronschweig: why do we generate a completely new http server instead of just using the web hook one that is started by the manager?

services/rebac-authz-webhook/pkg/config/config.go
```diff
@@ -70,6 +82,11 @@ func (cfg *Config) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&cfg.MetricsBindAddress, "metrics-bind-address", cfg.MetricsBindAddress, "Set the metrics bind address")
 	fs.StringVar(&cfg.HealthProbeBindAddress, "health-probe-bind-address", cfg.HealthProbeBindAddress, "Set the health probe bind address")
 	fs.StringVar(&cfg.OpenFGAAddr, "openfga-addr", cfg.OpenFGAAddr, "Set the OpenFGA address")
+	fs.StringVar(&cfg.BatchAuthzBindAddress, "batch-authz-bind-address", cfg.BatchAuthzBindAddress, "Set the HTTP bind address for batch authorization endpoint")
+	fs.DurationVar(&cfg.BatchAuthzReadTimeout, "batch-authz-read-timeout", cfg.BatchAuthzReadTimeout, "Set the read timeout for batch authorization HTTP server")
+	fs.DurationVar(&cfg.BatchAuthzWriteTimeout, "batch-authz-write-timeout", cfg.BatchAuthzWriteTimeout, "Set the write timeout for batch authorization HTTP server")
+	fs.DurationVar(&cfg.BatchAuthzIdleTimeout, "batch-authz-idle-timeout", cfg.BatchAuthzIdleTimeout, "Set the idle timeout for batch authorization HTTP server")
+	fs.DurationVar(&cfg.BatchAuthzShutdownTimeout, "batch-authz-shutdown-timeout", cfg.BatchAuthzShutdownTimeout, "Set the graceful shutdown timeout for batch authorization HTTP server")
```
aaronschweig: we would not need all these configurations if we just reuse the internal one from the manager

aaronschweig (changes requested):

services/rebac-authz-webhook/pkg/handler/batch/types.go
```diff
@@ -0,0 +1,58 @@
...
+
+import (
+	authorizationv1 "k8s.io/api/authorization/v1"
+)
+
+// BatchAuthzRequest is the input for the batch authorization endpoint.
+type BatchAuthzRequest struct {
+	// Items contains the list of authorization checks to perform.
+	Items []BatchAuthzItem `json:"items"`
+}
```
aaronschweig: why do we reinvent the wheel here? I would have imagined, we just accept a slice of `SubjectAccessReviewRequests` which is the standard k8s type for that. This way we could also just reuse our normal handler logic and need way less code changes in general

aaronschweig:
> for batch endpoint a new http server has been added to the webhook, because server from manager's webhook works as an https server. To be aligned with iam-service and graphql gateway we need http server


I also disagree with that statement, if we pass the proper certificates to the calling services, which should only be the portal for the batch endpoint as the rest is handled k8s internally, the internal https server is totally fine and IMHO the preferred option rather than rewiring the whole machinery

services/rebac-authz-webhook/cmd/serve.go
OlegErshov: it was done only to support current network configuration patterns. With a decision on handling https this server doesn't make sense anymore and I've removed it and used webhook's server

services/rebac-authz-webhook/pkg/config/config.go
OlegErshov: with webhook's server these configurations are not needed anymore, I've removed it

services/rebac-authz-webhook/pkg/handler/batch/types.go
OlegErshov: I've added a custom request struct to make it more simple for the frontend side, but it's also perfectly fine to use SARs to be more align with kubernetes. I've updated the expected request data and use a slice of SARs now.

The only change from original SAR handling is the cluster key from extra section. Kcp uses cluster names for authorization but to do the same, portal will need to extract cluster names from the accountInfo objects what adds more time complexity for preparing this batch-check. On the backed side we already have a cache with all needed data, so it will be more smooth to use cluster-path instead of the cluster-name in the request data. And make mapping `cluster path -> name `on the backend side

services/rebac-authz-webhook/pkg/handler/batch/batch.go
```diff
@@ -0,0 +1,213 @@
...
+		}
+
+		clusterInfo, ok := h.clusterCache.Get(clusterName)
+		if !ok {
+			h.log.Error(nil, "cluster info not found in cache", "clusterName", clusterName)
+			continue
+		}
+
+		if storeID == "" {
+			storeID = clusterInfo.StoreID
```
aaronschweig, line 182: buildChecks grabs storeID from the first SAR that resolves, but each SAR can carry a different clusterPath (→ different cluster → different store). Everything then gets sent to that one store:

```go
  if storeID == "" {
      storeID = clusterInfo.StoreID
  }
  // ...
  batchReq := &openfgav1.BatchCheckRequest{StoreId: storeID, Checks: checks}
```

If a batch mixes clusters, checks for the other clusters hit the wrong store, find no matching tuples, and come back allowed: false.

Can a single batch span multiple clusters? If not, let's assert all resolved store IDs match and return 400 on mismatch. If yes, we need to group checks by store and do one BatchCheck per store.

OlegErshov: Thank you, I've added defensive check to ensure that all clusters come from only one organization. All current use cases in portal will use clusters only from one org but if it changes in future we will get 400 error and change the behavior by grouping checks by store

aaronschweig (approved):

PR #222, opened 2026-07-29
feat: add search response totalcount

### LINKED ISSUE #359: feature: return total count in SearchResponse for page-based pagination
## Background

Related to #341 (page-based pagination support to `SearchRequest`).

When the backend implements page-based pagination, the `SearchResponse` must also return a `totalCount` so that consumers can render full pager controls (page numbers, first/last navigation, total items indicator).

## Problem

Currently `SearchResponse` returns a `NextCursor *string` for sequential traversal but no total count. With page-based pagination in place, the frontend still cannot render:

- Total number of pages (`ceil(totalCount / limit)`)
- First / last page navigation
- Total items indicator in the footer

## Proposed Response Shape

Based on the existing `SearchResponse` JSON convention (camelCase):

```go
type SearchResponse struct {
    Results    []SearchHit `json:"results"`
    NextCursor *string     `json:"nextCursor,omitempty"`
    TotalCount *int        `json:"totalCount,omitempty"`
}
```

`NextCursor` should only be populated when the request uses cursor-based pagination (i.e. `Cursor` is set as input). For page-based requests (`Page` is set), `NextCursor` should be omitted — the frontend can derive all pager state from `totalCount`, `page`, and `limit` alone.

| Request input | `NextCursor` in response | `TotalCount` in response |
|---|---|---|
| `Cursor` set | ✅ populated | optional but nice to have |
| `Page` set | ❌ optional but probably not requiered | ✅ required |

This keeps the two pagination modes cleanly separated and avoids the frontend needing to reconcile a cursor and a total count at the same time.

## Acceptance Criteria

- [ ] `SearchResponse` includes a `TotalCount *int` field when responding to page-based requests
- [ ] Proto / API contract updated if applicable
- [ ] Unit tests cover the `TotalCount` field in page-based responses

## Additional Context

Without `TotalCount`, the frontend correctly detects that the total is unknown and hides pager controls on non-last pages (current behavior after [openmfp/webcomponents#215](https://github.com/openmfp/webcomponents/pull/215)). Adding `TotalCount` to the response is the backend change needed to fully resolve the pagination UX.

pteich:
This PR adds a `totalcount` to the search response according to https://github.com/platform-mesh/backlog/issues/359

Page-based pagination scans and authorizes the full matching result set to return an exact authorized `totalCount`.

The latest changes now make use of a indexed account value that we can leverage to do a pre-filtering in OpenSearch for documents that a user is allowed to see.

The final OpenFGA batch check is still there as a last defense line. 

Fixes https://github.com/platform-mesh/backlog/issues/359


services/search-service/README.md
```diff
@@ -42,9 +42,12 @@ Query params:
...
 If both `page` and `cursor` are provided, cursor-based pagination takes precedence.
 
-Page-based pagination scans and authorizes preceding hits, so it is bounded by
-`MaxScannedHits / limit` (1000 scanned hits by default). Requests beyond that bound are rejected; use cursor-based
-pagination for deep traversal.
+Page-based pagination scans and authorizes the full matching result set to return an exact authorized `totalCount`.
+The scan is bounded by `MaxScannedHits` (1000 hits by default). If the service cannot exhaust the result set within
+that bound, it responds with `422 Unprocessable Entity`; use cursor-based pagination for broader or deeper traversal.
+
+TODO: Replace the exhaustive count with authorization-aware OpenSearch filters after access scopes such as workspace
```
lpgarzonr: Lets create an issue with details and more context and also add the issue link in the comments

pteich: the latest changes now use OpenSearch filtering, TODOs are removed

services/search-service/internal/clients/fga/authorizer.go
```diff
@@ -23,21 +23,65 @@ import (
...
 	openfgav1 "github.com/openfga/api/proto/openfga/v1"
+	"google.golang.org/grpc"
 
 	"go.platform-mesh.io/golang-commons/logger"
 	"go.platform-mesh.io/search-service/internal/service/search"
 )
 
-const batchCheckChunkSize = 100
+const (
+	batchCheckChunkSize   = 100
```
FWuermse: I remember FGA has some limit of 50 for batching. Do we already handle this further down the line? Maybe to be safe we can incorporate this hard limit of 50 already here.

services/search-service/internal/clients/fga/authorizer.go
```diff
@@ -23,21 +23,65 @@ import (
...
+	storeID, err := a.resolveStoreID(ctx, organization)
+	if err != nil {
+		return nil, fmt.Errorf("resolve store ID: %w", err)
+	}
+
+	res, err := a.client.ListObjects(ctx, &openfgav1.ListObjectsRequest{
+		StoreId:  storeID,
+		Type:     accountObjectType,
+		Relation: accountMemberRelation,
+		User:     fmt.Sprintf("user:%s", formatUser(user)),
```
FWuermse, line 58: If the user configures claim type email this will remove the dots in there. Is this also what is reflected in the FGA tuples when using email? I think so but not 100% sure anymore.

services/search-service/internal/clients/fga/authorizer.go
```diff
@@ -23,21 +23,65 @@ import (
...
 
 	"go.platform-mesh.io/golang-commons/logger"
 	"go.platform-mesh.io/search-service/internal/service/search"
 )
 
-const batchCheckChunkSize = 100
+const (
+	batchCheckChunkSize   = 100
+	accountObjectType     = "core_platform-mesh_io_account"
+	accountMemberRelation = "member"
```
FWuermse: This needs to be configurable. With the update of the provider config in the security operator it is possible to change the default role (which defaults to member). So let's take this from config.go.

FWuermse (commented):
Looks good! We just need to make the two consts configurable.

services/search-service/internal/clients/fga/authorizer.go
FWuermse: Hey, I just confirmed this by accident:

```sh
{"level":"error","service":"/Users/I762675/github.com/platform-mesh/platform-mesh/golang-
commons/logger/logger.go","rid":"b25ecaf2-7d26-429a-927e-77cc08566da0","error":"backend failure: 
filter authorization: openfga batch check: rpc error: code = Code(2000) desc = batchCheck received 100 
checks, the maximum allowed is 
50","path":"/rest/v1/search","organization":"sap","statusCode":500,"time":"2026-08-
18T15:22:50+02:00","caller":"/Users/I762675/github.com/platform-mesh/platform-mesh/services/search-
service/internal/router/router.go:108","message":"search request failed"}
```

pteich: done, added new config values

FWuermse (approved):

ntnn (approved):

PR #225, opened 2026-07-30
chore: search-config and field types in index

FWuermse:
Implements support for the search configuration as per [ADR-007](https://github.com/platform-mesh/architecture/blob/main/adr/007-search-config-resource.md) and ticket https://github.com/platform-mesh/backlog/issues/279.

Providers can configure which and how fields of their own APIResourceSchemas should be indexed. Default indexes all lexically.

On-behalf-of: https://github.com/SAP [florian.wuermseer@sap.com](mailto:florian.wuermseer@sap.com)

Also refactors responsibility of SearchIndex resource:
1. Binding Reconciler is now responsible for flattening hierarchy into SearchIndex
2. SearchIndex Reconciler replicates the field into Open Search
3. Indexing merely uses what Open Search configured for the given field

Fields from an APIResourceSchema can be:
- Made a filter (exact match)
- Made searchable via vector search (semantic search)
- Excluded from the Index completely

Example:

- Provider exports APIResourceSchema `apidefinitions`
- Provider has SearchConfig resource also named `apidefinitions`
- SearchConfig assigns field `spec.description` as semantic
- Open Search let's us perform vector search on that field
image


tjbutz (dismissed):
LGTM

pteich (dismissed):

hyperlingg (dismissed):

operators/search-operator/internal/config/config.go
```diff
@@ -43,7 +64,7 @@ type Config struct {
 		// IndexNamePrefix is a static prefix for all operator-managed index names and aliases.
 		IndexNamePrefix string `mapstructure:"opensearch-index-name-prefix" envconfig:"default=pm-orgs"`
 		// SemanticModelID is the OpenSearch ML model ID used for semantic field mappings.
-		SemanticModelID string `mapstructure:"opensearch-semantic-model-id" envconfig:"optional,OPENSEARCH_SEMANTIC_MODEL_ID"`
+		SemanticModelID string `mapstructure:"opensearch-semantic-model-id" envconfig:"optional,OPEN_SEARCH_SEMANTIC_MODEL_ID"`
```
lpgarzonr: Are u sure u want to modify that? i will expect opensearch to still be OPENSEARCH, if u decide to modified it please also update the rfc https://github.com/platform-mesh/architecture/blob/main/rfc/003-search-architecture.md?plain=1#L171

operators/search-operator/internal/config/config.go
```diff
@@ -30,7 +51,7 @@ type Config struct {
 	} `mapstructure:",squash"`
 
 	SearchableResource struct {
-		Resources []schema.GroupVersionKind `mapstructure:"resources" envconfig:"default={core.platform-mesh.io;v1alpha1;Account}"`
+		Resources []GroupVersionKindProvider `mapstructure:"resources" envconfig:"default={core.platform-mesh.io;v1alpha1;Account;platform-mesh-system}"`
```
lpgarzonr, line 54: This default is `;`-delimited, but `Unmarshal` splits on `,`. so parsing the
  default will probably fails since (`len(parts) != 4`) and the operator crashes on startup unless `RESOURCES` is overridden

operators/search-operator/internal/config/config.go
FWuermse: This env is just an internal convention and due to how `github.com/vrischmann/envconfig` resolved camel-case and for consistency `OPEN_SEARCH` would be the natural interpretation. However getting changes accepted in Architecture is harder than here so we should change it back.

operators/search-operator/internal/config/config.go
FWuermse: Unfortunately due to limits go struct tag syntax `github.com/vrischmann/envconfig` could not use `,` and instead uses a different syntax in the defaults and actual values. However, the interpretation will be parsed back so this ends up being a comma.

lpgarzonr (approved):

OlegErshov (approved):

operators/search-operator/internal/config/config.go
pteich: Agree, we should keep `OPENSEARCH_SEMANTIC_MODEL_ID`

PR #227, opened 2026-07-30
paginate ListStores when resolving FGA store

### LINKED ISSUE #226: FGA store lookup by name silently fails when store count exceeds one `ListStores` page
All FGA store-id-by-name resolution code calls `ListStores` with an empty request (no Name filter and no ContinuationToken) and then scans the returned slice in mem for a matching store name.

OpenFGA method `ListStores` is paginated and returns only a single, capped page per call with [some](https://openfga.dev/docs/interacting/read-tuple-changes#:~:text=The%20default%20page%20size%20is,does%20not%20expand%20the%20tuples.) default page size. So the lookup only ever inspects the first page of stores. Once the number of stores exceeds one page, any store on a later page is never found, since the lookup returns "store not found" even though the store exists.

nuromirg:
## Summary
Introduce a single shared helper (e.g. `store.FindStoreIDByName`) that paginates `ListStores` (using `PageSize` and `ContinuationToken`, and the server-side `Name` filter where an exact match is acceptable), and have all call sites use it instead of duplicating the ad-hoc first-page scan.

## Related issue(-s)
Fixes #226


nuromirg:
Hmm, ok, let me limit this PR with `golang-commons` changes first and then I'll create follow-up PR with operator/service separately.

xrstf (dismissed):

aaronschweig (approved):

PR #230, opened 2026-07-31
Add kro-composition-operator (KROaaS)

xmudrii:
## Add kro-composition-operator (KROaaS)

Adds a kcp-native composition operator that lets each tenant define their own composite APIs. A tenant writes a `ResourceGraphDefinition` (RGD) in their workspace; the operator turns it into a first-class API in that same workspace and materializes the composed children when instances are created.

### How

- Embeds kro's composition engine (`pkg/graph`, `pkg/runtime`, `pkg/dynamiccontroller`) as a library and drives it with a multicluster-runtime manager, watching RGDs through the `kro.run` APIExport virtual workspace.
- For each RGD it publishes the composite type as an APIResourceSchema in a per-RGD APIExport and self-binds it in the workspace. Being a bound resource is what makes the type first-class — the graphql-gateway schema and the security-operator OpenFGA model are both generated from APIBinding `boundResources`, so portal and authz work with no extra wiring.
- Teardown is finalizer-driven and ordered: the APIBinding is removed first (so the platform strips the type's authz) before the APIExport/APIResourceSchema.

### Why not kro or krop directly

- **kro** is single-cluster and not kcp-aware — no multicluster-runtime support, and none planned upstream right now. Running its engine across kcp workspaces needs MCR, hence embedding it as a library.
- **krop** closes the MCR gap too, but it's top-down: the provider defines composite types centrally. KROaaS is self-service — each tenant authors their own RGDs in their own workspace.

### adminAuth and the wildcard follow-up

The operator writes composed children (and manages the per-RGD APIExport/APIResourceSchema/APIBinding) directly into arbitrary consumer workspaces. To do that it uses its own kcp identity, deployed as a ManagedProvider with an `adminAuth: true` connection — i.e. cluster-admin — so nothing has to be provisioned per Account.

A per-workspace scoped identity was prototyped and dropped: PM's scoped mode derives its RBAC from the APIExport's permission claims, and kcp doesn't support wildcard permission claims — so a scoped identity can't be granted write access to the arbitrary child types a consumer-authored RGD composes.

The cost is that this identity is cluster-admin. Narrowing it to a scoped identity depends on kcp supporting wildcard permission claims; until then, adminAuth is required.

xref https://github.com/platform-mesh/backlog/issues/338


operators/kro-composition-operator/cmd/main.go
```diff
@@ -0,0 +1,175 @@
...
+
+	eng := engine.New(ctx, wsProvider, dynamiccontroller.Config{
+		Workers:              2,
+		ResyncPeriod:         10 * time.Hour,
+		QueueMaxRetries:      20,
+		MinRetryDelay:        200 * time.Millisecond,
+		MaxRetryDelay:        60 * time.Second,
+		RateLimit:            50,
+		BurstLimit:           100,
+		QueueShutdownTimeout: 30 * time.Second,
```
xrstf: Are any of these of importance, or could anyone change any of them? Like would it be bad to set RateLimit to 500 or 5? Can workers be reduced or increased here? Are we deviating from defaults here, and if so, why?

Just so we do not have 8 magic constants :)

operators/kro-composition-operator/cmd/main.go
```diff
@@ -0,0 +1,175 @@
...
+
+	reconLog := logf.Log.WithName("reconcile")
+	rec := mcreconcile.Func(func(ctx context.Context, req mcreconcile.Request) (ctrl.Result, error) {
+		err := eng.ReconcileRGD(ctx, string(req.ClusterName), req.Name)
+		if err == nil {
+			return ctrl.Result{}, nil
+		}
+		// Expected, self-resolving conditions requeue quietly; real failures are logged.
+		if engine.IsTransient(err) {
+			reconLog.V(1).Info("requeue (transient)", "cluster", string(req.ClusterName), "rgd", req.Name, "reason", err.Error())
```
xrstf, line 150: Isn't V1 like the fatal error shouting log level? I would have guessed transient errors should be on debug level, if at all, no?

operators/kro-composition-operator/internal/composition/contentconfig.go
```diff
@@ -0,0 +1,138 @@
...
+		query = "{ metadata { name namespace } }"
+		scope = "Namespaced"
+	}
+
+	listNode := map[string]any{
+		"pathSegment":             navSeg,
+		"navigationContext":       navSeg,
+		"label":                   kind + " (kro)",
+		"icon":                    "example",
+		"order":                   860,
```
xrstf: Is it expected that this number is 860 and in the other file it's 850?

operators/kro-composition-operator/internal/composition/graph.go
```diff
@@ -0,0 +1,79 @@
...
+func NewCompiler(cfg *rest.Config) (*Compiler, error) {
+	httpClient, err := rest.HTTPClientFor(cfg)
+	if err != nil {
+		return nil, fmt.Errorf("http client: %w", err)
+	}
+	b, err := graph.NewBuilder(cfg, httpClient)
+	if err != nil {
+		return nil, fmt.Errorf("graph builder: %w", err)
+	}
+	return &Compiler{builder: b}, nil
```
xrstf, line 57: This function should just call NewCompilerWithHTTPClient.

operators/kro-composition-operator/test/e2e/broader_test.go
```diff
@@ -0,0 +1,309 @@
...
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// TestMultiResourceComposition covers one RGD composing two distinct kinds
+// (ConfigMap + Secret) from a single instance, and instance-delete garbage
+// collection: deleting the instance collects every child via its owner refs.
+func TestMultiResourceComposition(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	t.Cleanup(cancel)
+
```
xrstf: Can this be t.Context()?

operators/kro-composition-operator/Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# NOTE: the build context is the repository ROOT, not this directory:
+#   docker build -f operators/kro-composition-operator/Dockerfile .
+# The module is linked through the repository-root go.work; this operator depends
+# only on external modules (kro, kcp SDK, controller-runtime), not the shared ./apis.
+FROM --platform=$BUILDPLATFORM golang:1.26 AS builder
```
xrstf: Should we do the same as all other Dockerfiles? ( https://github.com/platform-mesh/platform-mesh/blob/main/operators/security-operator/Dockerfile#L22 )

operators/kro-composition-operator/internal/composition/contentconfig.go
xmudrii: Yes, the other is for the kro provider itself (i.e. RGDs), these are for each RGD.

operators/kro-composition-operator/cmd/main.go
xmudrii: The values are now matching the kro default values.

operators/kro-composition-operator/test/e2e/broader_test.go
xmudrii: Fixed.

operators/kro-composition-operator/cmd/main.go
xmudrii: No, V1 is expected, logr is the other way around so V1 is debug logging.

operators/kro-composition-operator/Dockerfile
xmudrii: Good point, fixed.

operators/kro-composition-operator/internal/composition/graph.go
xmudrii: `NewCompilerWithHTTPClient` is not actually used at all so I ended up removing it.

xmudrii:
@xrstf All comments addressed, PTAL again.

operators/kro-composition-operator/cmd/main.go
xrstf: The joys of magic numbers, exhibit 14,382.

xrstf (approved):

Issue #104, opened 2026-06-09
T9: Cross-tier repair subroutine (orphan-tuple sweep, security-operator wait)


mjudeikis:
**Track:** Components
**Size:** M
**Epic:** platform-mesh/platform-mesh#99
**RFC:** [RFC 009 §Implementation Tickets — T9](https://github.com/platform-mesh/architecture/blob/main/rfc/009_backup-and-restore-operator.md#track-2--component-drivers-fully-parallel-after-track-1)
**Depends on:** platform-mesh/backup-operator#4

## Scope

After Velero restore completes, close the cross-tier boundary gap between kcp etcd and CNPG Postgres. Steps:

1. Wait for the security-operator to come up and re-reconcile `Store` / `AuthorizationModel` CRs (status-based wait, no direct OpenFGA calls from this subroutine).
2. List OpenFGA tuples whose `originClusterID/name` no longer resolves to a kcp resource on the target cluster.
3. Sweep them — `dry-run` by default; opt-in deletion via `PlatformRestore.spec.repair.delete`.
4. Report counts to `PlatformRestore.status.repair`.

No logic for Keycloak sessions, in-flight controller state, or anything outside this surface — those are explicitly out of scope.

## Deliverables

- [ ] `internal/repair/` subroutine
- [ ] Security-operator readiness wait (status-based)
- [ ] OpenFGA tuple listing + orphan detection by `originClusterID/name`
- [ ] Dry-run mode (default) and opt-in deletion flag wired through `PlatformRestore.spec.repair.delete`
- [ ] `status.repair` counts: scanned, orphaned, deleted

## Acceptance

- [ ] Dry-run reports a non-zero orphan count on a known-divergent target without mutating OpenFGA
- [ ] Opt-in deletion mode removes exactly the orphans reported by dry-run
- [ ] Subroutine never blocks waiting on a security-operator that is already up

PR #236, opened 2026-08-05
fix(security-operator): re-key orphaned account role tuples

### LINKED ISSUE #214: security-operator: role tuples are orphaned when a workspace's logical cluster changes; only the creator's tuples are re-created
## What happens

OpenFGA role tuples embed the logical-cluster hash of the workspace that owns the account, e.g. `role:core_platform-mesh_io_account/<clusterId>/<account>/<roleId>`. When a workspace is re-created (platform reinstall, disaster recovery, migration), the account gets a new logical cluster and every previously assigned role tuple keeps the old hash. iam-service resolves the clusterId live from AccountInfo per request, so the stale tuples don't error — the Members list just silently shrinks.

On workspace (re-)initialization, `AccountTuplesSubroutine` re-creates tuples for `Account.spec.creator` only (`operators/security-operator/internal/subroutine/account_tuples.go`). Members assigned later through iam-service exist solely as tuples under the old hash and are never re-created — the operator has no record of them.

The orphans also accumulate: the FGA store is adopted by name across workspace re-creations (`workspace_initializer.go` / `store.go` ListStores match-by-name), and tuple deletion only runs through the terminatingworkspaces provider on graceful termination. A wipe or hash change never matches the old keys.

## Observed in practice

On our Showroom dev environment (cc-d2), a workspace re-creation reduced an org's Members list to the creator alone; every other owner/member assignment was still in the store under the dead hash. We recovered by re-writing the roster against the current hash, but nothing warned us and nothing would have GC'd the dead tuples.

## Ask

Either tooling (re-key tuples on logical-cluster change, or an orphan GC/report), or an explicit statement in the docs that role assignments do not survive workspace re-creation and must be re-seeded.

anastazya:
Fixes #214.

When an org workspace is deleted and re-created under the same name, its logical cluster hash changes, but the OpenFGA role tuples keep the old hash in their object keys. The operator re-creates only the creator's tuples (`Account.spec.creator`), so every other member silently drops off the org roster — we hit this in production and had to reconstruct rosters by hand from the orphaned tuples (which survive in the store because it is adopted by name).

This adds an org-level scan-and-re-key subroutine to the OrgLogicalCluster controller, gated behind `--rekey-orphaned-tuples-enabled` (default off — without the flag, behavior is byte-identical):

- tuple filters are objectType-restricted so only role/org tuples for the reconciled org are touched (nothing else in a shared store)
- before re-keying, the old cluster is checked for liveness via a LogicalCluster GET; anything but NotFound/Forbidden fails closed (skip + requeue) so tuples of a live cluster are never moved
- writes-first-deletes-after, chunked at the 100-tuple write cap; a same-name descendant Account in the org workspace skips the entire pass (prevents cross-account key capture)
- emits event `OrphanedTuplesRekeyed` and metric `security_operator_rekeyed_tuples_total{result}`; idempotent on re-reconcile

Validation: all unit packages green (`go test ./internal/...`), table-driven tests for the filter/re-key/liveness/guard paths (+742 test lines); gofmt/gci clean. The kcp-envtest integration suite was not run locally (missing kcp binary) — CI covers it.

Related: #104 proposes a delete-only orphan sweep; note that deleting orphaned tuples destroys the only recoverable record of pre-re-creation rosters — re-keying preserves them. The flag default keeps this opt-in until operators choose the behavior.

### DIFF
--- a/operators/security-operator/internal/fga/rekey.go
+++ b/operators/security-operator/internal/fga/rekey.go
@@ -18,84 +18,9 @@ package fga
 
 import (
 	"slices"
-	"strings"
 
 	pmcorev1alpha1 "go.platform-mesh.io/apis/core/v1alpha1"
 )
-
-// AccountKey is the parsed form of the two FGA key shapes that embed an
-// account's origin cluster id:
-//
-//	role:<objectType>/<clusterID>/<name>/<role>[#<relation>]
-//	<objectType>:<clusterID>/<name>[#<relation>]
-//
-// When the org workspace backing an account is deleted and re-created, the
-// cluster id changes and every tuple built from these keys goes stale.
-type AccountKey struct {
-	ObjectType string
-	ClusterID  string
-	Name       string
-	Role       string // non-empty for the role shape
-	Relation   string // optional "#<relation>" suffix (e.g. "assignee")
-}
-
-// ParseAccountKey parses s into an AccountKey. It returns false for strings
-// that do not match either of the two shapes (e.g. "user:someone",
-// "role:authenticated#assignee").
-//
-// The match is purely structural: other key families share the raw
-// `<type>:<clusterID>/<name>` shape (e.g. the
-// `apis_kcp_io_apiexport:<providerClusterID>/<exportName>` user keys written
-// into the same org store for APIExportPolicies), so callers MUST additionally
-// check ObjectType against the configured account object type before treating
-// the result as an account key.
-func ParseAccountKey(s string) (AccountKey, bool) {
-	var k AccountKey
-
-	if i := strings.IndexByte(s, '#'); i >= 0 {
-		k.Relation = s[i+1:]
-		s = s[:i]
-		if k.Relation == "" {
-			return AccountKey{}, false
-		}
-	}
-
-	typ, rest, found := strings.Cut(s, ":")
-	if !found || typ == "" || typ == "user" {
-		return AccountKey{}, false
-	}
-
-	parts := strings.Split(rest, "/")
-	switch {
-	case typ == "role" && len(parts) == 4:
-		k.ObjectType, k.ClusterID, k.Name, k.Role = parts[0], parts[1], parts[2], parts[3]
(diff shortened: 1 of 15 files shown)


operators/security-operator/internal/fga/rekey.go
```diff
@@ -0,0 +1,149 @@
...
+//
+// When the org workspace backing an account is deleted and re-created, the
+// cluster id changes and every tuple built from these keys goes stale.
+type AccountKey struct {
+	ObjectType string
+	ClusterID  string
+	Name       string
+	Role       string // non-empty for the role shape
+	Relation   string // optional "#<relation>" suffix (e.g. "assignee")
+}
```
xrstf: This struct introduces the reverse implementation of the functions like `renderAccountEntity` or `RenderRolePrefix`. Is it now maybe time to put both implementation onto the same type (i.e. re-use AccountKey in tuples.go)? I fear this implementation will be able to drift away from the other, as  there is no mention of the other in either of them.

anastazya:
Good catch — fixed in b01bc71.

`AccountKey` now lives in `tuples.go` next to the renderers and owns the format in both directions: `renderAccountEntity`, `renderOwnerRole` and `renderOwnerRoleAssigneeGroup` build their keys through `AccountKey.String()`, and `RenderRolePrefix` delegates to a new `AccountKey.RolePrefix()`. `rekey.go` keeps only the re-key logic.

To keep them from drifting, `TestRenderersRoundTripThroughAccountKey` parses each renderer's output back and compares it against the fields it was built from — changing one direction without the other now fails the tests.

Also in 49d30e7: the events are emitted through `GetEventRecorder` instead of the deprecated `GetEventRecorderFor` (staticcheck SA1019 was failing `verify`). The new signature takes a `related` object and an `action`, so the action is now the constant `RekeyOrphanedTuples` while the reason keeps distinguishing the outcome; tests use `events.FakeRecorder`, which formats identically, so the assertions are unchanged. Plus two `intrange` findings in the tests.

xrstf (approved):

PR #237, opened 2026-08-05
add positioning paper on multi-region

### LINKED ISSUE #366: task: Write position paper on multi-region
### Overview

Summarize status quo, immediate next steps, and long term outlook

### Implementation Details

_No response_

### Completion Checklist

- [ ] Documentation updated
- [ ] Changes reviewed

SimonTheLeg:
Positioning paper on multi-region

fixes: https://github.com/platform-mesh/backlog/issues/366

### DIFF
--- a/contrib/poc/multi-region-position.md
+++ b/contrib/poc/multi-region-position.md
@@ -1,0 +1,140 @@
+# Multi-Region Status Quo and Position
+
+This paper explains the current status of multi-region support in platform-mesh. It does not touch on multi-zone support (see differentiation below).
+
+## Differentiation: Region vs Zone
+
+Regions are characterized by a significant geographical distance between them, resulting in round-trip times between 20-250ms. For example cross-continent, transatlantic or transpacific communication.
+
+Zones are geographically redundant locations with round-trip times between >1-20ms. For example locations on the same continent, with private backbones.
+
+## Status Quo - Global Control-Plane
+
+In order to run PlatformMesh across multiple regions acting as a global control-plane, it is split on the kcp level using shards. A kcp shard connects to its own etcd cluster and a centrally run CacheServer. Additionally shards communicate with each other during scheduling of logicalclusters. A shard can have multiple replicas, with one being active. Shards can be deployed as part of Kubernetes clusters or on standalone infrastructure. Shards can be deployed as part of Kubernetes clusters or on standalone infrastructure.
+
+Lastly a region can have multiple shards to extend its capacity horizontally.
+
+Here is a simplified example of a two-region setup, with the region `Europe` acting as the global control plane.
+
+```mermaid
+flowchart TB
+    subgraph America
+        subgraph USUserGroup["user"]
+            USUser(["us-user / controller"])
+        end
+        subgraph USPMesh["platform-mesh"]
+            subgraph US1Shard["kcp shard america1"]
+                US1Active["kcp replica (active)"]
+                US1Passive1["kcp replica (passive)"]
+                US1Passive2["kcp replica (passive)"]
+                US1VW["virtual-workspace"]
+            end
+            subgraph US2Shard["kcp shard america2"]
+                US2Active["kcp replica (active)"]
+                US2Passive1["kcp replica (passive)"]
+                US2Passive2["kcp replica (passive)"]
+                US2VW["virtual-workspace"]
+            end
+            subgraph US1Etcd["etcd cluster america1"]
+                US1Etcd1["etcd"]
+                US1Etcd2["etcd"]
+                US1Etcd3["etcd"]
+            end
+            subgraph US2Etcd["etcd cluster america2"]
+                US2Etcd1["etcd"]
+                US2Etcd2["etcd"]
+                US2Etcd3["etcd"]
+            end
+            US1Shard --> US1Etcd
+            US2Shard --> US2Etcd
+            US1Shard <-->|logicalcluster scheduling| US2Shard
+            USFP["frontproxy"]
+        end
+    end
+
+    subgraph Europe
+        subgraph EUUserGroup["user"]
+            EUUser(["eu-user / controller"])
+        end
+        subgraph EUPMesh["platform-mesh"]
+            subgraph EU1Shard["kcp shard europe1"]
+                EU1Active["kcp replica (active)"]
+                EU1Passive1["kcp replica (passive)"]
+                EU1Passive2["kcp replica (passive)"]
+                EU1VW["virtual-workspace"]
+            end
+            subgraph EU2Shard["kcp rootshard"]
+                EU2Active["kcp replica (active)"]
+                EU2Passive1["kcp replica (passive)"]
+                EU2Passive2["kcp replica (passive)"]
+                EU2VW["virtual-workspace"]
+            end
+            subgraph EU1Etcd["etcd cluster europe1"]
+                EU1Etcd1["etcd"]
+                EU1Etcd2["etcd"]
+                EU1Etcd3["etcd"]
+            end
+            subgraph EU2Etcd["etcd cluster europe2"]
+                EU2Etcd1["etcd"]
+                EU2Etcd2["etcd"]
+                EU2Etcd3["etcd"]
+            end
+            CacheServer["CacheServer"]
+            EUPortal["PlatformMesh portal"]
+            OpenFGA["OpenFGA"]
+            EUFP["frontproxy"]
+
+            EU1Shard --> EU1Etcd
+            EU2Shard --> EU2Etcd
+            EU1Shard <-->|logicalcluster scheduling| EU2Shard
+            EUFP -->|discovery / auth| EU2Shard
+        end
+    end
+
+    EU1Shard <--> CacheServer
+    EU2Shard <--> CacheServer
+    US1Shard <--> CacheServer
+    US2Shard <--> CacheServer
+    EU1Shard -->|authz check| OpenFGA
+    EU2Shard -->|authz check| OpenFGA
+    US1Shard -->|authz check| OpenFGA
+    US2Shard -->|authz check| OpenFGA
+    EUFP -->|auth / data-plane proxying| EU1Shard
+    EUFP -->|auth / data-plane proxying| EU2Shard
+    EUFP -->|auth / data-plane proxying| US1Shard
+    EUFP -->|auth / data-plane proxying| US2Shard
+    USFP -->|discovery / auth| EU2Shard
+    USFP -->|auth / data-plane proxying| EU1Shard
+    USFP -->|auth / data-plane proxying| EU2Shard
+    USFP -->|auth / data-plane proxying| US1Shard
+    USFP -->|auth / data-plane proxying| US2Shard
+    EUUser --> EUFP
+    EUUser --> EU1VW
+    EUUser --> EU2VW
+    EUUser --> US1VW
+    EUUser --> US2VW
+    EUUser --> EUPortal
+    USUser --> USFP
+    USUser --> EU1VW
+    USUser --> EU2VW
+    USUser --> US1VW
+    USUser --> US2VW
+    USUser --> EUPortal
+    EU1Shard <-->|logicalcluster scheduling| US1Shard
+    EU1Shard <-->|logicalcluster scheduling| US2Shard
+    EU2Shard <-->|logicalcluster scheduling| US1Shard
+    EU2Shard <-->|logicalcluster scheduling| US2Shard
+```
+
+## Resulting Next Steps
+
+The following improvements to the current architecture should be made in the mid-term to improve its viability for multi-region to minimize and streamline cross-region connections in a global-control plane model.
+
+* Build a distributed CacheServer
+* Run distributed OpenFGA or allow for Kubernetes RBAC to be configurable in platform-mesh
+* Make shards not talk to each other for scheduling logicalclusters
+* Run portal distributed across regions, with a shared database or create per region portals
+
+## Far Horizon
+
+Multi-region setups need to minimize the amount of traffic and requests made between regions. In the long-term the global control-plane should take in more of an indexing role, providing clients with information how to connect to all regional control planes, similar to the Cloud Hyperscaler model.


contrib/poc/multi-region-position.md
```diff
@@ -0,0 +1,134 @@
+# Multi-Region Status Quo and Position
+
+This paper explains the current status of multi-region support in platform-mesh. It does not touch on multi-zone support (see differentiation below).
+
+## Differentiation: Region vs Zone
+
+Regions are characterized by a significant geographical distance between them, resulting in round-trip times between 20-250ms. For example cross-continent, transatlantic or transpacific communication.
+
+Zones are geographically redundant locations with round-trip times between >1-20ms. For example locations on the same continent, with private backbones.
```
xrstf, line 9: Is there / what is the term for what AWS would call an Availability Zone, e.g. a different floor / building in the same data center? I fear "zone" might otherwise be confusing (it reads to me, if I map our names to AWS for example, eu-central and eu-west would be different _zones_ in our terminology).

contrib/poc/multi-region-position.md
```diff
@@ -0,0 +1,134 @@
...
+    USUser --> EUPortal
+    EU1Shard <-->|logicalcluster scheduling| US1Shard
+    EU1Shard <-->|logicalcluster scheduling| US2Shard
+    EU2Shard <-->|logicalcluster scheduling| US1Shard
+    EU2Shard <-->|logicalcluster scheduling| US2Shard
+```
+
+## Resulting Next Steps
+
+The following improvements to the current architecture should be made in the mid-term to improve its viability for multi-region:
```
xrstf: Based on what criteria exactly?

For example, do we want to allow users using the portal to navigate through the full workspace tree? Or should they only see the workspaces on a currently selected shard (similar to how existing cloud providers do it)?

Can users choose where (region/zone) to create workspaces (in PM speak, accounts)?

How do multiple shards per region work exactly, in terms of addressing them? If it's just a flat list of shards (8 in europe, 9 in asia, 3 in US), how much does regionality even affect PM (i.e. what is the difference between zones and regions, just the ping? Will a region become a zone if I upgrade to fiber?)

contrib/poc/multi-region-position.md
```diff
@@ -0,0 +1,134 @@
...
+
+Zones are geographically redundant locations with round-trip times between >1-20ms. For example locations on the same continent, with private backbones.
+
+## Status Quo
+
+In order to run PlatformMesh across multiple regions, it is split on the kcp level using shards. A kcp shard connects to its own etcd cluster and a centrally run CacheServer. Additionally shards communicate with each other during scheduling of logicalclusters. A shard can have multiple replicas, with one being active. Shards can be deployed as part of Kubernetes clusters or on standalone infrastructure. If Kubernetes clusters are chosen, we recommend to use one cluster per region.
+
+Lastly a region can have multiple shards to extend its capacity horizontally.
+
+Here is a simplified example of a two-region setup, with the region `Europe` acting as the global control plane.
```
xrstf, line 17: Ngl, routing every single request from US through an EU proxy feels ... like vindica.. I mean like a bottleneck. Is having a single front-proxy really viable? Or is the front-proxy DNS round-robbin'ed and runs once per region (or per zone)?

If we want users to explicitly choose what region they want to work with, we could have one FP per region. If truly the entire mesh should be behind 1 kubeconfig (i.e. one FP), .. toughie.

ntnn: I'd expect the front-proxy address in global setup to be on a BGP network, so they are automatically routed to the closes front-proxy.

contrib/poc/multi-region-position.md
ntnn: I agree, AZs are already prevalent in DCs and that could be confusing when talking about multi-region.

contrib/poc/multi-region-position.md
```diff
@@ -0,0 +1,134 @@
...
+
+## Differentiation: Region vs Zone
+
+Regions are characterized by a significant geographical distance between them, resulting in round-trip times between 20-250ms. For example cross-continent, transatlantic or transpacific communication.
+
+Zones are geographically redundant locations with round-trip times between >1-20ms. For example locations on the same continent, with private backbones.
+
+## Status Quo
+
+In order to run PlatformMesh across multiple regions, it is split on the kcp level using shards. A kcp shard connects to its own etcd cluster and a centrally run CacheServer. Additionally shards communicate with each other during scheduling of logicalclusters. A shard can have multiple replicas, with one being active. Shards can be deployed as part of Kubernetes clusters or on standalone infrastructure. If Kubernetes clusters are chosen, we recommend to use one cluster per region.
```
ntnn:
```suggestion
In order to run PlatformMesh across multiple regions, it is split on the kcp level using shards. A kcp shard connects to its own etcd cluster and a centrally run CacheServer. Additionally shards communicate with each other during scheduling of logicalclusters. A shard can have multiple replicas, with one being active. Shards can be deployed as part of Kubernetes clusters or on standalone infrastructure. 
```

I wouldn't say one cluster per region - on the contrary I'd more lean towards placing shards across AZs to make it less likely that a shard fails.

contrib/poc/multi-region-position.md
SimonTheLeg: This was chosen on purpose: Zones is the common denominator across hyperscalers. In the case of AWS they call it Availability Zones. But these are not different floor / building of the same datacenter, "Availability Zones in a Region are meaningfully distant from each other, up to 60 miles" (https://docs.aws.amazon.com/whitepapers/latest/aws-fault-isolation-boundaries/availability-zones.html). Zones are solely characterized by their low latency. For example in AWS it is even "single-digit millisecond latency", but that is not the same for all providers, so I have chosen >1-20 to capture roughly the estimations from all hyperscalers.

contrib/poc/multi-region-position.md
SimonTheLeg: The intent was to say, that we don't recommend spanning a single cluster across regions. I'll rework it a little to stress that point more.

I would not include any Zones into this. As this highly depends on your infrastructure provider. For example on AWS, if you have single digit latency between AZs, it can be perfectly viable to run your Kubernetes cluster across AZs, which is an option in EKS as well.

We should keep our recommendation as generic as possible here.

SimonTheLeg: a wait I possibly read this wrong. Okay makes sense now, and is even more generic

contrib/poc/multi-region-position.md
SimonTheLeg: alright, but nonetheless, two frontproxies. Will update the graph

contrib/poc/multi-region-position.md
SimonTheLeg: I have added some criteria, but the overall idea is a global control-plane (aka everyone can see everything) while trying to limit (but not yet fully optimize), the traffic between regions

> For example, do we want to allow users using the portal to navigate through the full workspace tree?

the full tree

> Can users choose where (region/zone) to create workspaces (in PM speak, accounts)?

I would say yes. Since it is a global control-plane and you don't connect to one region as of now. 

> how much does regionality even affect PM (i.e. what is the difference between zones and regions, just the ping? Will a region become a zone if I upgrade to fiber?

Yes. If by upgrading to fiber you mean dropping the latency below 20ms, then yes

contrib/poc/multi-region-position.md
```diff
@@ -0,0 +1,140 @@
+# Multi-Region Status Quo and Position
+
+This paper explains the current status of multi-region support in platform-mesh. It does not touch on multi-zone support (see differentiation below).
+
+## Differentiation: Region vs Zone
+
+Regions are characterized by a significant geographical distance between them, resulting in round-trip times between 20-250ms. For example cross-continent, transatlantic or transpacific communication.
```
mjudeikis, line 7: This is kinda misleading. Regions can be country-specific or just a legal boundary (UK -> France) close enough that they could be zones, but one could do regions. Its more about physical co-location. Not "let's cross the ocean".

contrib/poc/multi-region-position.md
mjudeikis: Right, same goes for the region. Zones are same location, under the same datacenter boundary (meaning where once can still run latency-sensitive services and be ok).

Feels like we talking about `Region` -> `Datacenter` (westeurope), `Zone` -> `Datacenter availability zone/Failure domain` (westeurope-1,2,3) ? Or is there a different understanding?

contrib/poc/multi-region-position.md
```diff
@@ -0,0 +1,140 @@
...
+
+Zones are geographically redundant locations with round-trip times between >1-20ms. For example locations on the same continent, with private backbones.
+
+## Status Quo - Global Control-Plane
+
+In order to run PlatformMesh across multiple regions acting as a global control-plane, it is split on the kcp level using shards. A kcp shard connects to its own etcd cluster and a centrally run CacheServer. Additionally shards communicate with each other during scheduling of logicalclusters. A shard can have multiple replicas, with one being active. Shards can be deployed as part of Kubernetes clusters or on standalone infrastructure. Shards can be deployed as part of Kubernetes clusters or on standalone infrastructure.
+
+Lastly a region can have multiple shards to extend its capacity horizontally.
```
mjudeikis, line 15: For me this breaks here. 
Set of shard represents Regions. So as 3 availability zone datacenter in westeurope you would have:
1. Shard1 - westeurope1
2. Shard2 - westeurope2
3. Shard3 - wsteurope3

FrontProxy would be one per 3 shards `westeurope`. So basically, the thing you see in the UI is `Region` westeurope, with 3 availability zones (distinct shards)?

contrib/poc/multi-region-position.md
mjudeikis: I think we need to be more specific. We could say - we get one instance of frontproxy per "region" (if we follow my suggestions on terminology). 

And the feature we need to work on is fallback of the frontproxy. Like some clouds, they can route you via different "front gates" if local is down. Same datacenter, but you enter to it via different datacenter front-gate

contrib/poc/multi-region-position.md
```diff
@@ -0,0 +1,140 @@
...
+
+Here is a simplified example of a two-region setup, with the region `Europe` acting as the global control plane.
+
+```mermaid
```
mjudeikis, line 19: as discussed, I think we need to drop openFGA, and leave comments on how some of the services must be handled. Like there will always be some sort of global services. 2 types comes to my mind:
1. Requires almost real-time convergence - authentication, authorization - you need access here and now. Still eventual consisten, but low latency
2. Requires eventual consistency - billing (usage summary can be aggregated as long as it fits into billing system window(usually 10-15min))

contrib/poc/multi-region-position.md
xrstf: > For example, do we want to allow users using the portal to navigate through the full workspace tree?

Does this mean the UI has to connect to every single shard in the system in order to build the workspace tree?

contrib/poc/multi-region-position.md
SimonTheLeg: Hmm so for both AWS and GoogleCloud westeurope1 and westeurope2 are the names of the regions, not westeurope. They then use a letter "a", "b", "c" to denote the zone, eg westeurope-1a, westeurope-1b, etc.

My intention is to side-step the terminology, which can be slightly unique to every provider and instead simply define the two terminologies based on the latency and nothing else.
From our perspective, we don't really care if it is in the same datacenter, one that is 10km, 100km or 1000km away as long as the latency meets our criteria of not being higher than 20ms. By doing it this way, I belive we also are able to help people with private datacenters to check how they should classify. their setup.

wdyt?

contrib/poc/multi-region-position.md
SimonTheLeg: yes in the current architecture this is case. Otherwise I don't see how - with things we currently have - we can create this global control-plane which is able to list all items.

From my point of view, this can only change if we implement what is hinted at in the "far horizon" section, where we would have some sort of indepentend synching mechanism for this type of global data, which then does not require the UI to connect to every shard.

contrib/poc/multi-region-position.md
```diff
@@ -0,0 +1,140 @@
...
+The following improvements to the current architecture should be made in the mid-term to improve its viability for multi-region to minimize and streamline cross-region connections in a global-control plane model.
+
+* Build a distributed CacheServer
+* Run distributed OpenFGA or allow for Kubernetes RBAC to be configurable in platform-mesh
+* Make shards not talk to each other for scheduling logicalclusters
+* Run portal distributed across regions, with a shared database or create per region portals
+
+## Far Horizon
+
+Multi-region setups need to minimize the amount of traffic and requests made between regions. In the long-term the global control-plane should take in more of an indexing role, providing clients with information how to connect to all regional control planes, similar to the Cloud Hyperscaler model.
```
mengliwg, line 140: In the future we might also consider another deployment model: independent regional Platform Mesh deployments with optional cross-region API sharing

contrib/poc/multi-region-position.md
xrstf: I would simply question why we want to have both a globally-distributed system, but also treat it as one global singleton. That feels like opposed ideas fighting with each other.

Can we / should we really make the UI handle all the eventualities? If one shard (region) is down, does the UI in every single request an every single UI element have to be prepared for incomplete data and retries and all that? Does this also mean the UI is always as slow as the slowest shard/region?

Feels like the backend team saying "let's spread the data across the globe" and then the UI team "why did you spread it all across the globe, now we need to collect it from everywhere just to show the UI.".

My 3 Pfennige: This is the wrong approach and will lead to endless frustration both for users and developers. At least I wouldn't want to be tasked with building a UI on top of this ^^

SimonTheLeg: Yeah I agree with those points. I think what we need to do as a project with this specific task is either a) accept these risks in order to build something with our current possibilites or b) extend the next step to include for a truly distributed control-plane

contrib/poc/multi-region-position.md
SimonTheLeg: I don't think I can follow atm.

In the current proposal there is an instance of frontproxy per region and each frontproxy connects to all shards, meaning we have the routing via different front gates if one of them is down.

What is missing from your point of view?

contrib/poc/multi-region-position.md
SimonTheLeg: I don't think I can quite follow: The terminology both in AWS and Google Cloud is:

westeurope1 and westeurope2 are the names of the regions, not westeurope. They then use a letter "a", "b", "c" to denote the zone, eg westeurope-1a, westeurope-1b, etc.

With this, are you referring to an architecture with

Region westeurope1
shard1 -> westeurope1a
shard2 → westeurope1b

Region westeurope2
shard1 -> westeurope2a
shard2 → westeurope2b

An then one frontproxy per region. With that being said, I think this is what is shown in the current proposal in the architecture diagram. Or am I missing something?

contrib/poc/multi-region-position.md
SimonTheLeg: Hmm. I see the point how it can be confusing since in hyperscaler lingo, "regions" can be close enough to each other (e.g. GC europe-west3 is Frankfurt and europe-west10 is Berlin) and therefore have sub 20ms latency and be a "zone" in our lingo.

We can rename this if we want to and use a term maybe "location", which google cloud is using for these. But then we also need to rename this whole paper, since I guess it is then about multi-location and not region anymore.

I wanted to sidestep this whole thing since I think when we talk about multi-region, we are talking about places for infrastructure that are far away from each other. But if that is not the case, then we need to actually go back to the requirements drawing board.

contrib/poc/multi-region-position.md
SimonTheLeg: Sorry I cannot follow: Discussed where and with whom?

I thought for now openFGA is required to run platform-mesh, it will not start without it right?

I have also put "Run distributed OpenFGA or allow for Kubernetes RBAC to be configurable in platform-mesh" as a Resulting Next step to address this issue for the future.

xrstf: > I thought for now openFGA is required to run platform-mesh, it will not start without it right?

Yes, but IMHO we should consider openFGA to be providing something that can be provided externally. Compare this to kcp, which is absolutely nothing you need to run yourself.

Or to put it differently: The diagram leaves out Keycloak, even though it currently also is a 100% necessary dependency of Platform Mesh. If Keycloak is "external", so should we view OpenFGA, too.

contrib/poc/multi-region-position.md
mengliwg: > For example, do we want to allow users using the portal to navigate through the full workspace tree? Or should they only see the workspaces on a currently selected shard (similar to how existing cloud providers do it)?

I would say it depends on the personas. Platform Administrator : full tree. Consumers: regions governed by policy, they can choose regions that are allowed for them.

SimonTheLeg:
proposal has been re-written according to our discussions. Ready for re-review

SimonTheLeg:
ptal, and approve if it looks good for you

PR #255, opened 2026-08-11
feat: add entityType validation for ContentConfiguration

### LINKED ISSUE #75: Add entityType Validation for Content Configuration processing in Operator and ContentConfig validation in Server
# Context

Currently users do not get any feedback if storing a ContentConfiguration that uses `entityType` referencing a non existing entityType.

# Target

To improve this we should validate the entityType value against all known entityTypes in the system.

- Process:
  - When reconciling a ContentConfiguration or validating one
  - Build map of all known entity types by parsing existing ContentConfigurations in the system and taking the `defineEntity` segments.
  - then validate against the user entityType and provide appropriate feedback
  - Consider global always a valid entityType.

akafazov:
Validate entityType references in ContentConfigurations against known entity types defined via `defineEntity` segments across all CCs in the system. Unknown entity types are reported as hard validation failures.

The feature is gated behind `--entity-type-validation-enabled` (default **false**) on both the operator and server.

Closes #75

---

Migrated from https://github.com/platform-mesh/extension-manager-operator/pull/702 — that PR was created before the monorepo existed and is now superseded by this one.

### DIFF
--- a/operators/extension-manager-operator/cmd/server.go
+++ b/operators/extension-manager-operator/cmd/server.go
@@ -21,7 +21,6 @@ import (
 	"encoding/json"
 	"errors"
 	"net/http"
-	"os"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus/promhttp"
@@ -33,8 +32,6 @@ import (
 	platformmeshcontext "go.platform-mesh.io/golang-commons/context"
 	"go.platform-mesh.io/golang-commons/traces"
 
-	"k8s.io/client-go/rest"
-	"k8s.io/client-go/tools/clientcmd"
 	ctrl "sigs.k8s.io/controller-runtime"
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 
@@ -116,9 +113,9 @@ func RunServer(_ *cobra.Command, _ []string) { // coverage-ignore
 func initServerEntityTypeRegistry(ctx context.Context) *validation.EntityTypeRegistry { // coverage-ignore
 	registry := validation.NewEntityTypeRegistry()
 
-	k8sClient := buildServerK8sClient()
-	if k8sClient == nil {
-		return registry
+	k8sClient, err := buildServerK8sClient()
+	if err != nil {
+		log.Fatal().Err(err).Msg("entity type validation is enabled but failed to connect to cluster")
 	}
 
 	loadRegistryFromCluster(ctx, k8sClient, registry)
@@ -127,30 +124,17 @@ func initServerEntityTypeRegistry(ctx context.Context) *validation.EntityTypeReg
 	return registry
 }
 
-func buildServerK8sClient() ctrlruntimeclient.Reader { // coverage-ignore
-	kubeconfigPath := os.Getenv("KUBECONFIG")
-	var restCfg *rest.Config
-	var err error
-	if kubeconfigPath != "" {
-		restCfg, err = clientcmd.BuildConfigFromFlags("", kubeconfigPath)
-		if err != nil {
-			log.Warn().Err(err).Msg("failed to load kubeconfig, entity type validation will only recognize 'global'")
-			return nil
-		}
-	} else {
-		restCfg, err = rest.InClusterConfig()
-		if err != nil {
-			log.Warn().Err(err).Msg("not running in cluster, entity type validation will only recognize 'global'")
-			return nil
-		}
+func buildServerK8sClient() (ctrlruntimeclient.Reader, error) { // coverage-ignore
+	restCfg, err := ctrl.GetConfig()
+	if err != nil {
+		return nil, err
 	}
 
 	k8sClient, err := ctrlruntimeclient.New(restCfg, ctrlruntimeclient.Options{Scheme: scheme})
 	if err != nil {
-		log.Warn().Err(err).Msg("failed to create k8s client, entity type validation will only recognize 'global'")
-		return nil
+		return nil, err
 	}
-	return k8sClient
+	return k8sClient, nil
 }
 
 func loadRegistryFromCluster(ctx context.Context, k8sClient ctrlruntimeclient.Reader, registry *validation.EntityTypeRegistry) { // coverage-ignore
--- a/operators/extension-manager-operator/cmd/server.go
+++ b/operators/extension-manager-operator/cmd/server.go
@@ -18,19 +18,25 @@ package cmd
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"net/http"
+	"os"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/spf13/cobra"
 
+	pmuiv1alpha1 "go.platform-mesh.io/apis/ui/v1alpha1"
 	"go.platform-mesh.io/extension-manager-operator/internal/server"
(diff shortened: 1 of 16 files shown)


operators/extension-manager-operator/pkg/validation/entitytype.go
```diff
@@ -0,0 +1,204 @@
...
+	}
+}
+
+// normalizeEntityType strips any "::" suffix (e.g. "project.overview::compound"
+// becomes "project.overview").
+func normalizeEntityType(entityType string) string {
+	if base, _, found := strings.Cut(entityType, "::"); found {
+		return base
+	}
+	return entityType
```
xrstf:
```go
base, _, _ := strings.Cut(entityType, "::")
return base
```

or even shorter via

```go
return strings.SplitN(entityType, "::", 2)[0]
```

operators/extension-manager-operator/pkg/validation/entitytype.go
```diff
@@ -0,0 +1,204 @@
...
+	if node.DefineEntity != nil && node.DefineEntity.Id != "" {
+		fullType := buildEntityTypeName(entityType, node.DefineEntity.Id)
+		types[fullType] = true
+	}
+
+	// Determine the entity type context for children
+	childEntityType := entityType
+	if node.DefineEntity != nil && node.DefineEntity.Id != "" {
+		childEntityType = buildEntityTypeName(entityType, node.DefineEntity.Id)
+	}
```
xrstf: Can we merge the two conditions, they seem identical. We could also reduce the repetitve object access like so:

```go
childEntityType := entityType
if entity := node.DefineEntity; entity != nil && entity.Id != "" {
  // Determine the entity type context for children
  fullType := buildEntityTypeName(entityType, entity.Id)
  types[fullType] = true

  childEntityType = buildEntityTypeName(entityType, entity.Id)
}

File: operators/extension-manager-operator/pkg/validation/entitytype.go [hunk 3]
```diff
@@ -0,0 +1,204 @@
+package validation
```
xrstf, line 17: Not sure why, but this file should make the boilerplate check job fail. It's missing the license header comment.

operators/extension-manager-operator/pkg/subroutines/contentconfiguration.go
```diff
@@ -125,9 +173,71 @@ func (r *ContentConfigurationSubroutine) Process(ctx context.Context, obj ctrlru
...
 
+func ownerKey(cc *pmuiv1alpha1.ContentConfiguration) string {
+	if cc.UID != "" {
+		return string(cc.UID)
+	}
+	return cc.Namespace + "/" + cc.Name
+}
+
+func (r *ContentConfigurationSubroutine) initEntityTypeRegistry(ctx context.Context, log *logger.Logger) error {
+	if r.k8sReader == nil {
```
xrstf: Instead of checking this here, late, why not check this in the New() function and prevent a broken subroutine object in the first place?

operators/extension-manager-operator/internal/server/validate_endpoint.go
```diff
@@ -124,3 +123,21 @@ func (h *HttpValidateHandler) HandlerValidate(w http.ResponseWriter, r *http.Req
...
+	var responseErr Response
+	for _, e := range merr.Errors {
+		responseErr.ValidationErrors = append(responseErr.ValidationErrors, validationError{
+			Message: e.Error(),
+		})
+	}
+
+	responseBytes, _ := json.Marshal(responseErr)
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusOK)
```
xrstf: Why are validation errors HTTP 200? I would have expected a 400 (Bad Request).

operators/extension-manager-operator/cmd/server.go
```diff
@@ -100,3 +112,80 @@ func RunServer(_ *cobra.Command, _ []string) { // coverage-ignore
...
+	}
+
+	loadRegistryFromCluster(ctx, k8sClient, registry)
+	log.Info().Int("entityTypes", len(registry.KnownTypes())).Msg("initialized server entity type registry")
+	go refreshRegistryPeriodically(ctx, k8sClient, registry)
+	return registry
+}
+
+func buildServerK8sClient() ctrlruntimeclient.Reader { // coverage-ignore
+	kubeconfigPath := os.Getenv("KUBECONFIG")
```
xrstf: This feels somehow very out of place for a Kubernetes operator to only rely on $KUBECONFIG and not offer a CLI flag to override it. IMHO we should either add our own flag or rely on client-go's magic (personally I'm no big fan of its flag handling).

operators/extension-manager-operator/cmd/server.go
```diff
@@ -100,3 +112,80 @@ func RunServer(_ *cobra.Command, _ []string) { // coverage-ignore
...
+	if kubeconfigPath != "" {
+		restCfg, err = clientcmd.BuildConfigFromFlags("", kubeconfigPath)
+		if err != nil {
+			log.Warn().Err(err).Msg("failed to load kubeconfig, entity type validation will only recognize 'global'")
+			return nil
+		}
+	} else {
+		restCfg, err = rest.InClusterConfig()
+		if err != nil {
+			log.Warn().Err(err).Msg("not running in cluster, entity type validation will only recognize 'global'")
```
xrstf: So when I supply a broken kubeconfig or my cluster is down, the operator will just emit a warning and continue? Wouldn't this mean I might never notice that it is not doing what I wanted it to do when I supplied the kubeconfig?

IMHO the operator should crash if it cannot connect to the cluster.

operators/extension-manager-operator/cmd/server.go
akafazov: Now using ctrl.GetConfig() - same patterns as other operators and also handles CLI flags.

operators/extension-manager-operator/cmd/server.go
akafazov: Now logs error and exists in case of broken kubeconfig.

operators/extension-manager-operator/internal/server/validate_endpoint.go
akafazov: Now returning

```
StatusUnprocessableEntity          = 422 // RFC 9110, 15.5.21
```

in case of validation errors.

operators/extension-manager-operator/pkg/subroutines/contentconfiguration.go
akafazov: Improved error handling in subroutine

operators/extension-manager-operator/pkg/validation/entitytype.go
akafazov: LICENSE added

operators/extension-manager-operator/pkg/validation/entitytype.go
akafazov: yes, good catch

operators/extension-manager-operator/pkg/validation/entitytype.go
akafazov: OK

xrstf (dismissed):

xrstf (approved):

Issue #220, opened 2026-07-28
T13: Implement an end-to-end `backup workflow` for the Platform Mesh using the `Backup-Operator`


Elbehery:
**Track:** PoC  
**Size:** L  
**Epic:** platform-mesh/platform-mesh#99

## Scope

Implement an end-to-end `backup workflow` for the Platform Mesh using the `Backup-Operator`.

A `PlatformBackup` resource coordinates the capture of the data and configuration required to recover a Platform Mesh installation. The workflow covers credentials, KCP/etcd state, PostgreSQL data, OpenFGA data, Kubernetes resources, and recovery-related KCP policies.

The backup operator executes the required capture subroutines, waits for asynchronous backup resources to complete, and records the resulting artefacts in `PlatformBackup.status`.

PoC simplifications:

- Platform component names and namespaces are fixed to the PoC deployment layout.
- CloudNativePG backup targets the fixed `platform-mesh-pg` cluster.
- Object-storage buckets and key layouts follow the PoC conventions.
- Per-component and per-cluster overrides are deferred.

## Components Covered

- Platform and application credentials
- Kcp  etcd root and shards
- CloudNativePG/PostgreSQL
- OpenFGA data
- Kcp authorization and recovery policies
- Kubernetes resources managed through Velero

## Deliverables

- [x] Implement the `PlatformBackup` reconciliation workflow.
- [x] Execute backup subroutines in the required order.
- [x] Respect the component enablement settings in `PlatformBackup.spec.components`.
- [x] Inventory and capture credentials required during platform recovery.
- [x] Discover the KCP etcd shards participating in the backup.
- [x] Trigger an etcd snapshot for every required shard.
- [x] Wait for every etcd snapshot to complete.
- [x] Record per-shard snapshot keys and timestamps in `PlatformBackup.status.artefacts.etcd`.
- [x] Create an on-demand CloudNativePG `Backup` for `platform-mesh-pg`.
- [x] Wait until the CNPG backup reports `Completed`.
- [x] Verify that the WAL segment required by the CNPG backup is present in object storage.
- [x] Record the CNPG backup resource name in `PlatformBackup.status.artefacts.cnpg.backups`.
- [x] Export the OpenFGA data required to restore authorization state.
- [x] Store the OpenFGA dump in object storage under the backup ID.
- [x] Capture the KCP policies and authorization resources required during recovery.
- [x] Store the KCP policy manifest in object storage.
[...]

PR #256, opened 2026-08-11
fix(gateway): clean up ClusterAccess schemas on deletion

### LINKED ISSUE #129: Gateway v2(BUG): Deleting ClusterAccess with custom spec.path leaves orphaned schema file
When a ClusterAccess resource that has a custom spec.path is deleted, the schema file written at the path-derived location is never cleaned up.

The deletion cleanup runs inside the NotFound branch of Reconcile — at that point the object is already gone from the API server, so spec.path is inaccessible. The code falls back to req.Name to derive the schema file path, which is only correct when spec.path is unset.

### Root Cause

Example: a CR named my-cluster with spec.path: "custom-path" writes its schema to single-custom-path, but on deletion the cleanup attempts to delete single-my-cluster — a file that doesn't exist — leaving single-custom-path orphaned.

### Impact
Orphaned schema files accumulate on disk (or in the configured schema handler backend). The gateway may continue serving a stale schema for a cluster that no longer has a corresponding ClusterAccess CR.

### Fix
Add a finalizer (gateway.platform-mesh.io/schema-cleanup) so the controller can read the full object — including spec.path — before it disappears. The NotFound fallback is retained for backward compatibility with CRs that predate the finalizer.

hellsontime:
## Summary

- add a finalizer so the effective ClusterAccess schema path remains available during deletion
- remove schemas using the same resolved path used during generation
- retain the finalizer and retry when backend cleanup fails
- allow deletion to complete when the schema is already absent
- preserve name-based cleanup for resources created before the finalizer existed

This builds on and hardens the approach proposed in the archived [kubernetes-graphql-gateway PR #220](https://github.com/platform-mesh/kubernetes-graphql-gateway/pull/220).

Fixes #129

## Testing

- `go test ./... -count=1`
- `task lint`
- focused envtest coverage for custom paths, default paths, missing schemas, and cleanup failure/recovery

### DIFF
--- a/services/kubernetes-graphql-gateway/listener/controllers/clusteraccess/clusteraccess_controller_test.go
+++ b/services/kubernetes-graphql-gateway/listener/controllers/clusteraccess/clusteraccess_controller_test.go
@@ -20,8 +20,11 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"os"
 	"path/filepath"
+	"slices"
+	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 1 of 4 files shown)


aaronschweig (dismissed):
approved but with minor minor improvement suggestion

services/kubernetes-graphql-gateway/listener/controllers/clusteraccess/clusteraccess_controller_test.go
```diff
@@ -49,15 +54,33 @@ import (
 	mcreconcile "sigs.k8s.io/multicluster-runtime/pkg/reconcile"
 )
 
-const testNamespace = "test-ns"
+const (
+	testNamespace          = "test-ns"
+	schemaCleanupFinalizer = "gateway.platform-mesh.io/schema-cleanup"
```
aaronschweig: [very nip] this is a duplicate of the constant already introduced in the `clusteraccess_controller.go` if we make the constant public we could reuse that instead of redeclaring it here.

hellsontime: Applied in 175500f7: exported `SchemaCleanupFinalizer` and reused it from the external test package. Thanks!

aaronschweig (approved):

Issue #241, opened 2026-08-07
Spike: explore options for MSP visibility permissions at the org level


brsmnv:
PM admins need to control which orgs can see MSP offerings and integrations.

Goal of this spike: verify the viability of an org-scoped visibility mechanism end to end and gather initial feedback.

## Context

Today only `APIExportPolicy` restricts which accounts can bind an export. A (usually - single) policy exists per APIExport and controls behavior at binding time.  

We need a separate mechanism focused on the org (identity-side). Currently visibility cannot be separated from the permission to bind. This new feature will cover the discovery requirements in https://github.com/platform-mesh/backlog/issues/364

Updated the issue with the content from the [last comment](https://github.com/platform-mesh/platform-mesh/issues/241#issuecomment-5330387396)

---

## Final (hopefully) design (3)

A new custom resource represents grants to see provider exports. The resource lives in the organizational hierarchy, and all child workspaces underneath receive the grant by association (ancestry). 

Currently only grants at a single canonical location (level in the tree) is supported. For PM this would default to `:root:orgs:<*>`.

The implementation would be similar to #273 (just less code).

Edit: work on (diffed against 273) https://github.com/platform-mesh/platform-mesh/pull/288

### Scope

VisibilityGrants control the visibility and discovery of marketplace listings. Once granted, the set of visible APIExports cannot be scoped down at a child node deeper in the workspace hierarchy. 

A visible APIExport is not automatically a bindable export. APIExportPolicies are the provider-side authorization mechanism at binding time. 

Marketplace listings are not filtered by policies today, and reusing & overloading them with per-organization visibility control was decided against. We are intentionally splitting "can see" from "can bind" at the resource level.
#### Custom resource
Custom resource (high level):
- Providers are referenced by their logical cluster IDs.
- The provider exports are a plain list of APIExport names (strings).
- The subject of the grant:
	- is not on the spec itself;
	- is determined by the location of the grant resource;

Without the grant, the VW service defaults to a deny and the organization marketplace remains empty.

### Permissions and RBAC

Only identities with certain permissions can create the grant resource. This is gated by two mechanisms, both based on Kubernetes RBAC. 

**Target workspace role/role binding**

Permissions for the target workspace (organization) must be granted to the designated market / platform administrators group. 

Alternatively, RBAC for the target workspace can be bypassed by going through the Marketplace/VG export's virtual workspace. PM / Marketplace content admins need a role binding granting them write permissions on the `apiexports/content` subresource.

[...]

PR #257, opened 2026-08-12
fix(virtual-workspaces): scope marketplace exports to the providers cluster

brsmnv:
When selecting the APIExports for the marketplace entry, we filter  by the content-for label:
- APIExport.labels["content-for"] == provider.meta.name

Given two APIExports in different workspaces (A and B) having:
- different names, specs;
- the same "content-for" label value

Listing the APIExports by provider A returns both A and B APIExports.

This results is a second entry for the B APIExport under the provider A ProviderMetadata entry. The marketplace would then show B's export as part of the provider A entry.

Fix / missing criteria:
- Match APIExports by the cluster ID from the annotation;
- Filter out unrelated APIExports.

**Note:** The first commits included some light refactoring to enable testing. I reverted some changes to the code to make sure im fixing an existing bug. 

The test reproducing the bug is `TestMarketplace_ExportsFromForeignCluster`.

Edit: related to #241

### DIFF
--- a/services/virtual-workspaces/pkg/marketplace/server.go
+++ b/services/virtual-workspaces/pkg/marketplace/server.go
@@ -33,6 +33,8 @@ import (
 	"k8s.io/apimachinery/pkg/util/yaml"
 	"k8s.io/apiserver/pkg/authorization/authorizer"
 	genericapiserver "k8s.io/apiserver/pkg/server"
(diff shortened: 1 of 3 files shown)


mjudeikis (approved):
looks bit odd, but I think its right

PR #259, opened 2026-08-12
refactor(search-operator): remove hardcoded OpenSearch credentials

akafazov:
## What

Align the search-operator with platform-mesh standards for operator configuration and startup.

- Introduce a `cmd/` package (`root.go` + `operator.go`) with a cobra command tree
- Replace stdlib `flag` + `envconfig` with `cobra`/`pflag` + `golang-commons/config.CommonServiceConfig`
- Replace envconfig-tagged nested `Config` struct with a flat `OperatorConfig` using `AddFlags`
- Remove hardcoded default values for OpenSearch connection parameters
- OpenSearch credentials (`URL`, `Username`, `Password`) remain env vars — they are secrets
- Replace `NewClientFromEnv` (which redundantly called envconfig and re-read env vars) with `NewConfigFromEnv` + `NewClientFromConfig`
- Remove the `vrischmann/envconfig` dependency

## Why

Every other operator in the platform uses the golang-commons cobra/pflag pattern for configuration. The search-operator was using a bespoke stdlib `flag` + envconfig approach that diverged from the standard, and had hardcoded credential defaults (`admin`/`admin`) that are unsuitable for production.


aaronschweig (dismissed):
everywhere else we use the new flag setup from golang-commons, it properly does not fit the scope of the PR, but the search-operator should definitely be in line with the platform-mesh standards

akafazov:
I've refactored the config to use the golang-commons as the other operators.

akafazov:
superseded by https://github.com/platform-mesh/platform-mesh/pull/346

PR #262, opened 2026-08-13
chore: discover schemas from bound exports per provider

FWuermse:
This change retires the provider configuration and discovers schemas and providers by the bindings in the workspaces of an org.

- APIBindingWatcher does not rely on config anymore
- Config does not need the Provider name/ws anymore
- Only bound resources are staged for indexing
- Backwards compatible in regards to index naming

Addresses [PR 370](https://github.com/platform-mesh/backlog/issues/370)

### DIFF
--- a/operators/search-operator/internal/subroutine/apibinding_watcher.go
+++ b/operators/search-operator/internal/subroutine/apibinding_watcher.go
@@ -139,11 +139,13 @@ func (s *apiBindingWatcherSubroutine) Process(ctx context.Context, instance runt
 		providerPath := b.Spec.Reference.Export.Path
 		providerClient, err := GetScopedClient(mcMngr.GetConfig(), mcMngr.GetScheme(), providerPath)
 		if err != nil {
+			log.Warn().Str("provider", providerPath).Msg("failed to get provider client")
 			continue
 		}
 
 		schemas, err := boundSchemasForBinding(ctx, providerClient, b)
 		if err != nil {
+			log.Warn().Str("binding", b.Name).AnErr("due to", err).Msg("failed to get bound Resource Schemas for binding")
 			continue
 		}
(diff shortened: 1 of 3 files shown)


operators/search-operator/internal/subroutine/apibinding_watcher.go
```diff
@@ -132,13 +130,37 @@ func (s *apiBindingWatcherSubroutine) Process(ctx context.Context, instance runt
...
+	for _, b := range otherBindings.Items {
+		if b.Status.Phase != kcpapisv1alpha1.APIBindingPhaseBound {
+			continue
+		}
+		mcMngr := s.mgr.GetLocalManager()
+
+		providerPath := b.Spec.Reference.Export.Path
+		providerClient, err := GetScopedClient(mcMngr.GetConfig(), mcMngr.GetScheme(), providerPath)
+		if err != nil {
+			continue
```
pteich, line 143: Is it ok to just continue here? consider at least log the error so that we can debug issues.

operators/search-operator/internal/subroutine/apibinding_watcher.go
```diff
@@ -132,13 +130,37 @@ func (s *apiBindingWatcherSubroutine) Process(ctx context.Context, instance runt
...
+		}
+
+		schemas, err := boundSchemasForBinding(ctx, providerClient, b)
 		if err != nil {
-			return ctrl.Result{}, errors.NewOperatorError(fmt.Errorf("resolve fields for binding %q: %w", binding.Name, err), true, false)
+			continue
```
pteich, line 149: same here, consider at least log it and then continue

pteich (dismissed):

OlegErshov (approved):

PR #264, opened 2026-08-13
chore: update README and diagrams for Search Operator

FWuermse:
Updates readme to reflect recent changes in [ADR 007](https://github.com/platform-mesh/architecture/blob/main/adr/007-search-config-resource.md) and [RFC 003](https://github.com/platform-mesh/architecture/blob/main/rfc/003-search-architecture.md).


aaronschweig (approved):
maybe we can have the source of the SVG as part of that as well for future editing/adjustments?

PR #268, opened 2026-08-14
feat(search): make JWT user claim configurable

### LINKED ISSUE #371: story: Search reflects provider permissions and user claim config
### User Story

As a Platform Mesh admin
I want Search to reflect provider permissions and user claim configuration
so that Search behaves consistently with the auth webhook.

# Context

The default user claim (E-Mail, ID, etc.) can be configured in the [security operator](https://github.com/platform-mesh/platform-mesh/blob/main/operators/security-operator/internal/config/config.go#L101). Also which FGA roles are default for each Resource in Platform Mesh can be configured via the provider permission claims (e.g. whether `member` is the default role or `authenticated_user`). 

# Proposed solution

- Introduce a new config field matching the Search Operator's config `user-claim` and extract this field from the JWT token
- Keep `email` the default `user-claim` if not configured
- Adjust and possibly rename the exact field `fga_account_object` to whatever is the default role for visibility

### Acceptance Criteria

- [ ] Search results represent a subset of what the auth webhook would show
- [ ] Correct FGA tuple is extracted 

### Parent Epic

_No response_

### Additional Context

_No response_

hellsontime:
## Summary

- add configurable `--user-claim` support to Search Service, defaulting to `email`
- retain arbitrary JWT claims in `golang-commons` through a dedicated `ClaimSet`
- use the configured claim as the OpenFGA user instead of the previous `mail`/`sub` fallback
- reject missing, blank, or non-string configured claims with `401 Unauthorized`
- document that the value must match the Security Operator configuration

Closes platform-mesh/backlog#371

## Testing

- `go test -count=1 ./...` in `golang-commons`
- `go test -count=1 ./...` in `services/search-service`
- `task lint` in both modules
- manually tested signed RS256 JWTs with `email`, `mail`, `sub`, `preferred_username`, and a custom claim
- verified missing and non-string claims return `401`
- started the real Search Service binary with different `--user-claim` values and verified requests pass or fail at the expected middleware stage

## Notes

The existing OpenFGA `get` relation is unchanged. This change only makes the JWT claim used as the user identity configurable.


services/search-service/internal/middleware/orgauth.go
```diff
@@ -100,11 +102,8 @@ func (o *OrgContextMiddleware) SetRequestContext() func(http.Handler) http.Handl
 				}
 			}
 
-			user := strings.TrimSpace(token.Mail)
-			if user == "" {
-				user = strings.TrimSpace(token.Subject)
-			}
-			if user == "" {
+			user, found := token.Claims.String(o.userClaim)
+			if !found || strings.TrimSpace(user) == "" {
```
xrstf: Is it intended to keep spaces in claimes like `"foobar "`? From reading the code it seems like we generally want to get rid of / ignore whitespace, but we only do it to detect entirely empty values? I'd do

```go
// no need to check for existence, the empty value "" suits us just fine
user := strings.TrimSpace(token.Claims.String(o.userClaim))
if user == "" { .... }
```

pteich: agree, we should trim ws in any case

hellsontime: good catch, and no it was not intentional. The previous mail/subject path returned a trimmed value as well.

pteich (dismissed):
We originally had a simpler approach in mind (just use subject or email) but this more general approach is a better choice imho.

xrstf (dismissed):

aaronschweig:
@hellsontime would be great if you could resolve the conflicts then we can merge the PR 💪

OlegErshov (approved):

Issue #207, opened 2026-07-21
Separate kubeconfigs per control-plane for resource-broker


ntnn:
Currently resource-broker accepts three kubeconfigs, the default `-kubeconfig`, the `-kcp-kubeconfig` (which hosts the APIExports and the coordination APIs) and the `-compute-kubeconfig` where Migration resources are created.

This needs more separation to act as a provider in PM. The `-kubeconfig` can stay as it is as the default kubeconfig for the mcmanager.
Add a `-coord-kubeconfig` purely for the coordination API.
Add a `-verification-kubeconfig` (or similar) for the verification tree. It should point at the workspace the `-verification-tree-root` points at.
Add a `-staging-kubeconfig` (or similar) for the staging tree. It should point at the workspace the `-staging-tree-root` point at.

PR #269, opened 2026-08-15
feat: separate kubeconfigs for resource-broker

OlegErshov:
On-behalf-of: @SAP aleh.yarshou@sap.com

This pr addresses this issue https://github.com/platform-mesh/platform-mesh/issues/207

 **Goal**: 

Allow the resource-broker to use separate kubeconfigs for different kcp workspace trees (coordination, verification,
  staging) instead of a single shared kubeconfig.

  **Changes**:
  - Added 3 new CLI flags: `-coord-kubeconfig`, `-verification-kubeconfig`, `-staging-kubeconfig`
  - Split the single `WorkspaceClientFunc` into purpose-specific client factories
  - Each falls back to `-kcp-kubeconfig` if not provided (backward compatible)

### DIFF
--- a/operators/resource-broker/pkg/broker/broker.go
+++ b/operators/resource-broker/pkg/broker/broker.go
@@ -86,6 +86,18 @@ type Options struct {
 	// Required.
 	ComputeConfig *rest.Config
 
+	// CoordConfig is the base config for the coordination workspace.
+	// If nil, derived from KcpConfig.
+	CoordConfig *rest.Config
+
+	// VerificationConfig is the base config for the verification tree.
+	// If nil, uses KcpConfig.
+	VerificationConfig *rest.Config
+
+	// StagingConfig is the base config for the staging tree.
+	// If nil, uses KcpConfig.
+	StagingConfig *rest.Config
+
 	// AcceptAPIName is the name of the APIExportEndpointSlice serving
 	// AcceptAPIs. All other APIExportEndpointSlices in the broker workspace
 	// serve brokered resources.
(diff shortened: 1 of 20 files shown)


operators/resource-broker/pkg/broker/broker.go
```diff
@@ -86,6 +86,18 @@ type Options struct {
...
+	// If nil, derived from KcpConfig.
+	CoordConfig *rest.Config
+
+	// VerificationConfig is the base config for the verification tree.
+	// If nil, uses KcpConfig.
+	VerificationConfig *rest.Config
+
+	// StagingConfig is the base config for the staging tree.
+	// If nil, uses KcpConfig.
+	StagingConfig *rest.Config
```
xrstf, line 99: Nitty, but it says "uses" for verification and staging, but "derived" for coord. This made me think that coord is special and get some extra tweaking, but in the code further down, all three configs default to kcpConfig in the same way.

`s/derived from/uses/g`  😁

operators/resource-broker/pkg/controller/coordbroker/stagingworkspace/binding_ready.go
```diff
@@ -53,7 +53,7 @@ func (s *bindingReadySubroutine) Process(ctx context.Context, obj ctrlruntimecli
 	}
 
 	wsPath := s.opts.StagingTreeRoot + ":" + sw.Name
```
xrstf, line 55: Since I've seen this a bunch of times now by scrolling through this diff, mostly in tests, let me just scream into the empty forest:

> Let's not manually construct workspace paths, but use the logicalcluster APIs instead. I know `x + ":" + y` works and is unlikely to ever break, but deep down, it irks me 😅 

Thanks. No need to now change anything about this existing [IMHO] blemish in this PR. ^^

PR #272, opened 2026-08-17
Fix failing provider deletion

iakmc:
## Summary
- Provider workspace deletion could hang forever: AuthorizationModel's fga-tuples finalizer looks up its Store to clean up FGA tuples, and if the Store is already gone it kept erroring and retrying instead of finishing
- Now treats a NotFound Store as "already cleaned up" and lets the finalizer clear

## Reproduction
- Reproduced: the permanent hang. Built a real provider (via provider-quickstart) + real consumer account, bound them, then orphaned the consumer's Store so the AuthorizationModel's StoreRef pointed at nothing. Provider deletion got stuck in "Deleting" for 13+ min straight with zero recovery, confirmed via security-operator logs showing the exact `Store ... not found` error repeating on every retry.
- Not reproduced: the "platform console becomes unresponsive" part from the original report. Portal endpoint stayed responsive (200 OK, ~10-15ms) the whole 13 min window in every test. Might need more scale (many stuck objects) or a different code path than what I tested - still open.

## Test plan
- [x] added unit test covering the NotFound path (asserts OpenFGA is never called)
- [x] full security-operator test suite passes
- [x] reproduced the hang on a local kind cluster (orphaned Store, provider delete stuck 13+ min with no recovery), rebuilt+redeployed with the fix, same scenario now deletes in ~10s

### DIFF
--- a/operators/security-operator/internal/subroutine/tuples.go
+++ b/operators/security-operator/internal/subroutine/tuples.go
@@ -28,6 +28,7 @@ import (
 	"go.platform-mesh.io/security-operator/internal/fga"
 	"go.platform-mesh.io/subroutines"
 
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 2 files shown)


xrstf (approved):

mjudeikis:
Sounds like we masking other issue for racy component 🤷 
/approved

PR #273, opened 2026-08-17
spike(vw): Marketplace visibility grant

brsmnv:
Filter marketplace entries based on resources living in consumer/tenant workspaces. 

Permissions are derived directly from spec/state and no reconciling and status patching for the rersolved cluster ID is necessary.

The implementation on this branch is a bit too complex and involves a naive tree walking up to root. The final design and implementation will include a simplified version of that with canonical location for an account's grants. 

Related: #241

PR #274, opened 2026-08-18
fix(marketplace)!: Migrate marketplaceentry APIExport to v1alpha2 in vw service

brsmnv:
This PR migrates the MarketplaceEntry `apiExport` field to the v1alpha2 type. The change fixes the broken CRD and ARS; types are regenerated. 

> (!!!) This is a breaking change for any `spec.apiExport` consumers. Field `latestResourceSchemas` is replaced with `resources`; `all` is dropped. We need to ensure that the UI is not relying on any of these.

Frontend PR: https://github.com/platform-mesh/marketplace-ui/pull/187

## Description
The MarketplaceEntries CRD (and ARS) fails validation on apply. The `apiExport` field references an older version of the KCP APIExport type. That type declares that `group` is the map key for permissionClaims, but it is neither made required, nor is a default provided. 

This is fixed in a newer version of the APIExport type. I couldn't find a related issue in KCP, only this PR which adds the default value to the v1alpha2 schema: https://github.com/kcp-dev/kcp/pull/3486
- link to exact line: https://github.com/kcp-dev/kcp/pull/3486/files#diff-db1d09cd3ca584946616da9eee212c73381cf0a6aa69d926d811b2630409fe04R637

The failing `x-kubernetes-list-map-keys` validation requires either a default value or a mandatory field.
- upstream docs: https://kubernetes.io/docs/reference/kubernetes-api/apiextensions/custom-resource-definition-v1/
> x-kubernetes-list-map-keys annotates an array with the x-kubernetes-list-type `map` by specifying the keys used as the index of the map. [snip] The properties specified must either be required or have a default value, to ensure those properties are present for all list items.

**Why is this bug suddenly a problem?**
Until now we constructed MarketplaceEntries in the VW service in-process, and returned them to the client. MarketplaceEntry objects do not exist anywhere in kcp / etcd state. In #273 I am adding VisibilityGrants (a persisted type) in the same API group, so the two marketplace ARS(chemas) and the APIExport are created for the first time in PM. Applying them failed and exposed the problem.

I am not sure why this isn't breaking kcp as v1alpha1 types are being served today. kcp doesnt use the ARS->CRDs flow? 

## Spec changes

APIExport field changes:
- `spec.latestResourceSchemas []string` replaced with `spec.resources` (object list)
- `permissionClaims[].all` replaced (see `defaultSelector`)
- NEW: `permissionClaims[].verbs` (required, minItems 1)
- Minor / fixes:
  - `permissionClaims[].group` has a `default: ""` added (the ARS bugfix)
  - `permissionClaims[].identityHash` same, default added

### DIFF
--- a/apis/marketplace/v1alpha1/marketplaceentry_types.go
+++ b/apis/marketplace/v1alpha1/marketplaceentry_types.go
@@ -21,7 +21,7 @@ import (
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
-	kcpapisv1alpha1 "github.com/kcp-dev/sdk/apis/apis/v1alpha1"
+	kcpapisv1alpha2 "github.com/kcp-dev/sdk/apis/apis/v1alpha2"
 )
 
 // MarketplaceEntrySpec defines the desired state of MarketplaceEntry.
@@ -35,7 +35,7 @@ type MarketplaceEntrySpec struct {
 	ProviderMetadata pmuiv1alpha1.ProviderMetadata `json:"providerMetadata"`
 
 	// PermissionClaims are the permission claims associated with the marketplace entry.
-	APIExport kcpapisv1alpha1.APIExport `json:"apiExport"`
+	APIExport kcpapisv1alpha2.APIExport `json:"apiExport"`
 }
 
 // MarketplaceEntryStatus defines the observed state of MarketplaceEntry.
--- a/apis/marketplace/v1alpha1/marketplaceentry_types.go
+++ b/apis/marketplace/v1alpha1/marketplaceentry_types.go
@@ -34,7 +34,7 @@ type MarketplaceEntrySpec struct {
 	// ProviderMetadata contains metadata about the provider of the marketplace entry.
 	ProviderMetadata pmuiv1alpha1.ProviderMetadata `json:"providerMetadata"`
 
-	// PermissionClaims are the permission claims associated with the marketplace entry.
+	// APIExport associated with the marketplace entry.
 	APIExport kcpapisv1alpha2.APIExport `json:"apiExport"`
 }
 
--- a/apis/marketplace/v1alpha1/roundtrip_fuzz_test.go
+++ b/apis/marketplace/v1alpha1/roundtrip_fuzz_test.go
@@ -50,13 +50,14 @@ var marketplaceEntrySeeds = []struct {
 			"apiExport": {
 				"metadata": {"name": "test-api-export", "annotations": {"kcp.io/path": "root:providers:foo"}},
 				"spec": {
-					"latestResourceSchemas": ["v260623-482e10b2.widgets.foo.io"],
+					"resources": [{"group": "foo.io", "name": "widgets",
+					 "schema": "v260623-482e10b2.widgets.foo.io", "storage": {"crd": {}}}],
 					"identity": {"secretRef": {"name": "widgets-identity", "namespace": "kcp-system"}},
 					"maximalPermissionPolicy": {"local": {}},
 					"permissionClaims": [
-						{"group": "", "resource": "secrets", "all": true, "identityHash": "abc123"},
-						{"group": "apps", "resource": "deployments",
-						 "resourceSelector": [{"name": "one", "namespace": "two"}]}
+						{"group": "", "resource": "secrets", "verbs": ["*"], "identityHash": "abc123"},
+						{"group": "apps", "resource": "deployments", "verbs": ["get", "list"],
+						 "defaultSelector": {"matchAll": true}}
 					]
 				}
 			}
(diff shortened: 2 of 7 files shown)


apis/marketplace/v1alpha1/marketplaceentry_types.go
```diff
@@ -21,7 +21,7 @@ import (
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
-	kcpapisv1alpha1 "github.com/kcp-dev/sdk/apis/apis/v1alpha1"
```
mjudeikis, line 24: I think we need to update this in all codebase at the same time, including rest of the resources. this might break things :/

brsmnv: Is publishing v1alpha2.MarketplaceEntry together with the kcpv1alpha2.APIExport change marginally better or reven recommended here?

brsmnv:
WRT breaking consumers: 
The marketplace UI is not referencing the one removed `all` field, but types and queries include it (e.g. the graphql query). So a minor change there is necessary for example. 
Other downstream components might be impacted as well but I couldn't find anything else.

mjudeikis:
oh. sad. I suspect we might need to make changes in a better-planned manner.

apis/marketplace/v1alpha1/marketplaceentry_types.go
brsmnv: Update here, in most places we are using v1alpha2. Only Marketplace references alpha1, and a few of the newer operators reintroduced the reference by accident.

brsmnv:
Only impacted component is the UI and we can prepare a forwards compatible release as we go. Ready for review.

anastazya:
One gap worth a conscious call here or in a follow-up: a claim that originated as v1alpha1 with resourceSelector reads back through v1alpha2 as `verbs: ["*"]` with the name/namespace restriction only in the compat annotation, so a UI composing bindings from `spec.apiExport` would broaden it. Fine to declare those exports unsupported, but better said out loud. Happy to donate the embed consistency test from #303 if useful.

apis/marketplace/v1alpha1/marketplaceentry_types.go
```diff
@@ -35,7 +35,7 @@ type MarketplaceEntrySpec struct {
 	ProviderMetadata pmuiv1alpha1.ProviderMetadata `json:"providerMetadata"`
 
 	// PermissionClaims are the permission claims associated with the marketplace entry.
```
xrstf: This comment is describing a field that doesn't exist. ^^

brsmnv: Good catch I'll clean it up :D

brsmnv: Fixed

brsmnv:
> One gap worth a conscious call here or in a follow-up: a claim that originated as v1alpha1 with resourceSelector reads back through v1alpha2 as `verbs: ["*"]` with the name/namespace restriction only in the compat annotation, so a UI composing bindings from `spec.apiExport` would broaden it. Fine to declare those exports unsupported, but better said out loud. Happy to donate the embed consistency test from #303 if useful.

I think I get your point, but does this then impact existing APIExports in e.g. test systems? From my understanding, with APIExport storage version being v1alpha2 (kcp), conversion to alpha1 is done on the fly. And default selector (see below) ends up nil etc.

This is an assumption on my side so correct me if im wrong (!), but with resourceSelector gone, relying on the remaining (already present) fields is not enough. And `defaultSelector` should be set so it does not default to matching all resources.

v1alpha2 export:
```
// defaultSelector is the default selector to use when creating APIBindings
// via WorkspaceType's defaultAPIBindings. If not set, the APIBinding will
// default to matchAll: true.
```
So in the end this all boils down to two things, one of which can be a bit annoying to deal with:

1) Update acceptedPermissionClaims in the UI client

The easy part: a hardcoded matchAll at the moment, we can read the correct information from the new APIExport type:
```ts
// graphql.service.ts
    const acceptedPermissionClaims =
      entry.spec?.apiExport?.spec?.permissionClaims?.map((claim) => {
        return {
          state: 'Accepted',
          group: claim.group ?? '',
          resource: claim.resource,
          identityHash: claim.identityHash,
          verbs: ['*'],
          selector: {
            matchAll: true,
            // ...
```

We need to instead infer this from the defaultSelector or adjust this in general. Same for the verbs, * being hardcoded. Can come straight straight from the claim verbs.

2) Updating APIExports wherever necessary

This is the more annoying part and I have no idea what exists out there in the wild. But describing in v1 terms while storing v2 (I mean v1alpha2) is bound to cause issues anyway.

So here any APIExports need to scope by label and permission claims sneed to be tightened. If these were created with the v1alpha1 model in mind, that might require a  bit of rethinking.

brsmnv:
Closing and continuing in #337 (branch on fork, can't resolve conflicts here).

PR #277, opened 2026-08-18
kro-composition-operator: replace the materializer with kro's instance controller

xmudrii:
Instance reconciliation now runs kro's own instance controller (`pkg/controller/instance`), one per consumer workspace and GVR. The hand-written materializer is deleted.

kro's controller takes every dependency as a parameter, so it runs unmodified across workspaces. The fan-out stays in this operator. It brings server-side apply with ApplySet pruning, reverse-topological deletion, kro's instance finalizer, readyWhen gating, instance status and conditions, Events, and the `kro.run/reconcile: suspended` annotation.

Other kro code replacing local equivalents:

- `resourcegraphdefinition.NewConditionsMarkerFor` for RGD conditions. This also produces the aggregate Ready condition and prunes the condition kro renamed in v0.9, neither of which the hand-written version did.
- `metadata.CompareRGDOwnership` for ownership comparison.
- `graph/crd/compat.CompareVersions` for schema compatibility.
- `pkg/features` for kro's feature gates, exposed through a new `--feature-gates` flag.

New behaviour:

- A schema change that breaks the one already served is refused, using kro's rules and its `kro.run/allow-breaking-changes` opt-out. The published schema keeps serving.
- An RGD claiming a group and kind another RGD already publishes is refused, naming the holder.
- An RGD that does not compile reports `GraphAccepted=False` with the build error rather than carrying no status at all.
- `status.resources` reports each node's resolved dependencies.
- `spec.schema.metadata` labels and annotations land on the published APIResourceSchema.
- Leader election takes its lease from the host cluster. Pointed at the kcp provider workspace it silently started zero controllers.


operators/kro-composition-operator/internal/engine/publish.go
```diff
@@ -159,12 +181,31 @@ func teardownComposite(ctx context.Context, c ctrlruntimeclient.Client, rgd *kro
...
+	// 4. Delete the identity secret kcp created for this APIExport. kcp does not
+	//    owner-reference it, so neither GC nor the RGD's owner refs collect it, and since a
+	//    composite APIExport is minted per RGD the consumer workspace would otherwise keep
+	//    one secret for every RGD ever created. Done last: the export is gone by now, so
+	//    nothing can still be serving from this identity.
+	if deleteIdentitySecret {
+		secret := &corev1.Secret{ObjectMeta: metav1.ObjectMeta{
+			Name:      identitySecret.Name,
+			Namespace: identitySecret.Namespace,
+		}}
```
xrstf: ctrlruntimeclient.ObjectKeyFromObject(identitySecret)

operators/kro-composition-operator/cmd/main.go
```diff
@@ -96,6 +102,22 @@ func main() {
...
+	krometrics.Register(crmetrics.Registry)
+
+	// The lease namespace is this pod's, which exists in the host cluster and not in a kcp
+	// workspace, so leader election has to point at the host cluster.
+	var leaderCfg *rest.Config
+	if leaderElect {
+		cfg, err := rest.InClusterConfig()
+		if err != nil {
+			setupLog.Error(err, "in-cluster config for leader election")
+			return
```
xrstf: Is this causing a clean shutdown (exit code 0) or an error that would be visible as a CrashLoop?

operators/kro-composition-operator/internal/composition/contentconfig.go
```diff
@@ -32,10 +33,30 @@ const AccountEntity = "core_platform-mesh_io_account"
...
+	if group == "" {
+		return fmt.Errorf("API group is empty")
+	}
+	for _, label := range strings.Split(group, ".") {
+		if !graphQLName.MatchString(label) {
+			return fmt.Errorf(
+				"API group %q cannot be served to the portal: the segment %q is not a valid GraphQL name "+
+					"(letters, digits and underscores only, not starting with a digit). Hyphens are the "+
+					"usual cause, so use something like %q instead",
+				group, label, "vault.demo.io")
```
xrstf, line 52: This example value does not pass the regex, since it contains dots.

xrstf: Nevermind.

operators/kro-composition-operator/internal/engine/engine.go
```diff
@@ -311,44 +350,92 @@ func (e *Engine) ensureController(clusterName string, wc *workspace.Clients) *ws
...
-		wc, err := e.Workspaces.For(ctx, clusterName)
-		if err != nil {
-			return err
+// drainInstances deletes every instance of gvr and reports whether they are all gone.
+// Runs before the type is unserved, so kro's cleanup still can.
+//
+// force stops waiting when the workspace is terminating, which would otherwise block
+// workspace and account deletion.
+func drainInstances(ctx context.Context, dyn dynamic.Interface, gvr schema.GroupVersionResource, force bool) (bool, error) {
+	list, err := dyn.Resource(gvr).List(ctx, metav1.ListOptions{})
```
xrstf: Could this possibly be optimized by doing a metadata-only list?

operators/kro-composition-operator/internal/engine/engine.go
```diff
@@ -311,44 +350,92 @@ func (e *Engine) ensureController(clusterName string, wc *workspace.Clients) *ws
...
+	// Deletes are issued. kro's instance reconciler still has to delete the children and
+	// release its finalizer, so report not-drained and let the caller requeue. In force mode
+	// nothing holds the instances any more, so do not hold the RGD either.
+	return force, nil
+}
 
-		inst, err := wc.Dynamic.Resource(gvr).Namespace(req.Namespace).Get(ctx, req.Name, metav1.GetOptions{})
+// releaseInstanceFinalizer strips kro's finalizer, which nothing can process any more.
+func releaseInstanceFinalizer(ctx context.Context, ri dynamic.ResourceInterface, name string) error {
+	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
xrstf, line 416: We technically already fetched (listed) the instance, could we short-circuit this by checking HasInstanceFinalizer() before doing the whole Retry/Get/Update loop? Otherwise we re-fetch every single object when force=true.

operators/kro-composition-operator/internal/engine/publish.go
```diff
@@ -159,22 +182,231 @@ func teardownComposite(ctx context.Context, c ctrlruntimeclient.Client, rgd *kro
...
+	if into == nil {
+		into = make(map[string]string, len(from))
+	}
+	for k, v := range from {
+		if _, taken := into[k]; !taken {
+			into[k] = v
+		}
+	}
+	return into
+}
```
xrstf, line 232:
```go
func mergeMissing(into, from map[string]string) map[string]string {
	out := maps.Clone(from)
	maps.Copy(out, into)

	return out
}
```

is a bit backwards, but maybe better? Not sure. Just wanted to present the stdlib alternative :D

xrstf: Not sure if the side effect was intended in your function though. I would avoid it, I experienced a lot of pain modifying labels and then realizing the same map was used in 3 places in an object and I accidentially changed all 3.

xrstf (approved):

PR #280, opened 2026-08-19
fix: error on mismatch between Search batch size and default openFGA batch size limit

### LINKED ISSUE #360: task: Bug when searching for Components
### Overview

# Searching for Entities other than Account fails

Currently only the queries for types Account return a valid response. Other resources fail in a error code 500 with no further information.

## Steps to reproduce

1. Send a non-empty query with `?resource=components`
2. Observe the HTTP response

### Implementation Details

_No response_

### Completion Checklist

- [ ] Documentation updated
- [ ] Changes reviewed

FWuermse:
Fixes [360](https://github.com/platform-mesh/backlog/issues/360)

### DIFF
--- a/services/search-service/cmd/server.go
+++ b/services/search-service/cmd/server.go
@@ -94,7 +94,7 @@ var serverCmd = &cobra.Command{
 		svc := searchservice.NewService(
 			searchIndexResolver,
 			openSearchClient,
-			fgaclient.NewAuthorizer(fgaClient),
+			fgaclient.NewAuthorizer(fgaClient, *serviceCfg),
 			metrics,
 			searchservice.ServiceConfig{
 				DefaultLimit:   serviceCfg.Search.DefaultLimit,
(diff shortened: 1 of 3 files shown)


Elbehery (dismissed):

aaronschweig:
This can also be adjusted by the `maxChecksPerBatchCheck` flag for openfga.

 I think this should not be a fixed value in search, as it is not a fixed value in openfga

FWuermse:
> This can also be adjusted by the `maxChecksPerBatchCheck` flag for openfga.
> 
> I think this should not be a fixed value in search, as it is not a fixed value in openfga

Is this unlimited when configured? Then we could have the batch size configurable and use the same value to configure openFGA.

aaronschweig:
> > This can also be adjusted by the `maxChecksPerBatchCheck` flag for openfga.
> > I think this should not be a fixed value in search, as it is not a fixed value in openfga
> 
> Is this unlimited when configured? Then we could have the batch size configurable and use the same value to configure openFGA.

No, the default (unconfigured) limit is 50.

[See this](https://raw.githubusercontent.com/openfga/openfga/refs/tags/v1.18.3/.config-schema.json)

```json
"maxChecksPerBatchCheck": {
            "description": "The maximum number of tuples allowed in a BatchCheck request.",
            "type": "integer",
            "default": 50,
            "x-env-variable": "OPENFGA_MAX_CHECKS_PER_BATCH_CHECK"
        },
```

but I would like to see this value in sync between the two components

aaronschweig (approved):

PR #292, opened 2026-08-20
Fix for race condition between AuthorizationModel and Store on delete

iakmc:
## Summary
Follow up on the [previous fix](https://github.com/platform-mesh/platform-mesh/pull/272) - that one just made the stuck deletion recoverable, but MJ was right that it didn't fix the actual race.

Problem: Store checks if anything still uses it before deleting itself, but that check can be outdated by the time it actually deletes. So it can delete while something new just started depending on it.

Fix: before an AuthorizationModel gets created, it now has to register itself on the Store first, using a finalizer. Kubernetes won't let you add a finalizer to something that's already being deleted, it just rejects it. So now if the Store is going away, the bind fails right there instead of racing it. Store's own check is now just reading its own finalizers, no more querying across workspaces.

## Fixes
[#2421](https://github.com/platform-mesh/helm-charts/issues/2421)

## Test plan
- [x] new unit tests for the register/deregister part
- [x] updated existing tests where the call sequence actually changed, didn't weaken any of them
- [x] ran the real e2e suite (actual kcp, no mocks) - this caught 2 real bugs during dev, both fixed
- [x] checked directly against a live API server that the finalizer rejection actually happens
- [x] reproduced the race condition pre-fix , 0 orphans after the fix across repeated attempts, on a clean cluster rebuilt from scratch

### DIFF
--- a/operators/security-operator/internal/subroutine/authorization_model_generation_test.go
+++ b/operators/security-operator/internal/subroutine/authorization_model_generation_test.go
@@ -103,12 +103,7 @@ func TestAuthorizationModelGeneration_Process(t *testing.T) {
 			mockSetup: func(manager *mocks.MockManager, lister *mocks.MockLister, cluster *mocks.MockCluster, kcpClient *mocks.MockClient) {
 				manager.EXPECT().ClusterFromContext(mock.Anything).Return(cluster, nil)
 				cluster.EXPECT().GetClient().Return(kcpClient)
-				kcpClient.EXPECT().Get(mock.Anything, mock.Anything, mock.Anything).RunAndReturn(func(ctx context.Context, nn types.NamespacedName, o ctrlruntimeclient.Object, opts ...ctrlruntimeclient.GetOption) error {
-					if _, ok := o.(*pmcorev1alpha1.AccountInfo); ok {
-						return nil
-					}
-					return nil
-				}).Once()
+				kcpClient.EXPECT().Get(mock.Anything, mock.Anything, mock.Anything).Return(nil).Once()
 				manager.EXPECT().GetCluster(mock.Anything, mock.Anything).Return(cluster, nil)
 				kcpClient.EXPECT().Get(mock.Anything, mock.Anything, mock.Anything).Return(assert.AnError)
 			},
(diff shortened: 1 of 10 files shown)


operators/security-operator/internal/subroutine/authorization_model_generation_test.go
```diff
@@ -149,6 +149,14 @@ func TestAuthorizationModelGeneration_Process(t *testing.T) {
 					}
 					return nil
 				}).Once()
+				// registerAuthorizationModelWithStore: fetch the Store, add its finalizer.
+				kcpClient.EXPECT().Get(mock.Anything, mock.Anything, mock.Anything).RunAndReturn(func(ctx context.Context, nn types.NamespacedName, o ctrlruntimeclient.Object, opts ...ctrlruntimeclient.GetOption) error {
+					if _, ok := o.(*pmcorev1alpha1.Store); ok {
+						return nil
+					}
+					return nil
```
xrstf: I know this is copied from the other tests, but... what is this doing exactly? Is it boilerplate?

Should we maybe have a helper like

```go
func NopGetter(ctx context.Context, nn types.NamespacedName, o ctrlruntimeclient.Object, opts ...ctrlruntimeclient.GetOption) error {
    return nil
}
```

and then do `kcpClient.EXPECT().Get(mock.Anything, mock.Anything, mock.Anything).RunAndReturn(NopGetter).Done()` everywhere?

But again, I am not familiar with mocking too much, but what is this line saying? "When calling .Get() on the kcpClient, run this function instead, and make sure it's just called once"? If so, how is this implementation compatible with controller-runtime? Isn't this then returning a nil err from the `.Get()` call (indicating success), but the given object `o` is never actually updated?

operators/security-operator/internal/subroutine/tuples_test.go
```diff
@@ -402,18 +402,75 @@ func TestTupleFinalizationWithAuthorizationModel(t *testing.T) {
...
 					store := o.(*pmcorev1alpha1.Store)
 					*store = pmcorev1alpha1.Store{
+						ObjectMeta: metav1.ObjectMeta{
+							Finalizers: []string{"authzmodel.core.platform-mesh.io/8a5edab282632443219e051e4ade2d1d5bbc671c781051bf1437897cbdfea0", "core.platform-mesh.io/fga-store"},
+						},
 						Status: pmcorev1alpha1.StoreStatus{
 							StoreID:              "store-id",
 							AuthorizationModelID: "auth-model-id",
 						},
 					}
```
xrstf, line 417: Here, this forexample, *this* I understand, how the callback is updating `o`.

xrstf (approved):

PR #295, opened 2026-08-21
Percent-encode colons in resource names

### LINKED ISSUE #293: Fix authorization to allow colons in resource names
Currently if the user tries to create a resource with a ":" in the name OpenFGA is not able to answer this.

Image

After talking with Aaron there is already a mechanism to handle colons in user names but it should also be handled for resource names, since ":" is actually pretty common there.

BergCyrill:
This PR fixes #293.

OpenFGA requires an object key to contain exactly one colon (separator).
Decided to go with percent-encoding instead of using the same ":" -> "." substitution currently in-place for users because this could collide and handle authorization requests for different resources incorrectly, e.g. "system:controller:foo" and "system.controller.foo" are both valid ClusterRoles.

I have only improved rebac-authz-webhook on this because there are no tuples so far that contain a colon in the resource name (OpenFGA would reject it). Colon-free resource names are ensured to encode to themselves. So there is no need to change existing object keys or tuples.

The user substitution could (or should) be handled the same but this will introduce a breaking change that will need some rekeying mechanism and also I tend to keep PRs small, focused and reviewable.

### DIFF
--- a/services/rebac-authz-webhook/pkg/util/name.go
+++ b/services/rebac-authz-webhook/pkg/util/name.go
@@ -1,0 +1,46 @@
+/*
+Copyright The Platform Mesh Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+import "strings"
+
+// nameEncoder is a strings.Replacer that percent-encodes the characters that
+// OpenFGA forbids in the identifier portion of an object key.
+var nameEncoder = strings.NewReplacer(
+	"%", "%25",
+	":", "%3A",
+)
+
+// EncodeName percent-encodes the characters that OpenFGA forbids in the
+// identifier portion of an object key.
+//
+// OpenFGA requires an object to contain exactly one colon, the separator
+// between type and identifier, so a Kubernetes resource name that itself
+// contains a colon (e.g. the ClusterRole "system:controller:foo") produces an
+// object that the server rejects outright.
+//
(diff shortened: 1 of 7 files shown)


services/rebac-authz-webhook/pkg/util/name.go
```diff
@@ -0,0 +1,40 @@
...
+// both-legal resource names and must not map onto the same key. '%' is encoded
+// first so the transformation stays injective for callers outside Kubernetes;
+// a Kubernetes name can never contain '%' itself.
+//
+// Names that contain neither character are returned unchanged, so keys that
+// OpenFGA already accepts today keep their current form.
+func EncodeName(name string) string {
+	name = strings.ReplaceAll(name, "%", "%25")
+	return strings.ReplaceAll(name, ":", "%3A")
+}
```
aaronschweig, line 46: [nip]

if we do a replacer instead of two individual `strings.ReplaceAll` we can perform the replacement in one pass instead of two. It should be negligible for smaller strings, but in case we need to extend the replacements for other reserved characters we would not need to append another `ReplaceAll` call for every replacement intent

```go
var nameEncoder = strings.NewReplacer(
      "%", "%25",
      ":", "%3A",
  )

  func EncodeName(name string) string {
      return nameEncoder.Replace(name)
  }
```

aaronschweig (dismissed):
Just a small nitpicky comment 😉 

Also lets make sure to properly update the entity that writes into FGA, for me it is also fine if that is put into a ticket for now, so somebody else can pick that up

services/rebac-authz-webhook/pkg/util/name.go
BergCyrill: Sounds good, will change this

BergCyrill:
> Just a small nitpicky comment 😉
> 
> Also lets make sure to properly update the entity that writes into FGA, for me it is also fine if that is put into a ticket for now, so somebody else can pick that up

Let's create a new issue for that. Afaik there aren't currently any resource-name-based writes.

services/rebac-authz-webhook/pkg/util/name.go
BergCyrill: done - I like this solution even more because it is ordering-independent. Thanks for your feedback

aaronschweig (approved):

PR #298, opened 2026-08-24
feat(kubernetes-graphql-gateway): add per-ClusterAccess schema filtering

### LINKED ISSUE #197: Allow ClusterAccess to restrict which resource types are exposed in the generated GraphQL schema
## Summary

Today, when the listener generates a GraphQL schema for a cluster, it exposes **every** resource type discovered from that cluster's OpenAPI v3 spec. There is no way to constrain the generated schema to a curated subset of types.

I'd like to request the ability to configure, per `ClusterAccess`, an **allowlist of resource types** that should be included in the generated schema. Only the selected types would get queries, mutations, and subscriptions generated for them; everything else would be omitted from the schema entirely.

## Motivation / use case

We want to use the gateway to expose a **public-facing API** built from a deliberately small, stable set of resource types, while the underlying cluster contains many more types that are internal and should never appear in that public schema.

Right now our only options are:
- expose the entire cluster surface (leaks internal types into a public API, and bloats the schema), or
- run a separate cluster whose CRD set exactly matches the desired API (heavy, and not always feasible).

A schema-level type filter on `ClusterAccess` would let a single gateway serve a **narrow, intentional public API** over a cluster that has a broader set of resources. The schema becomes a contract we control by configuration rather than by whatever happens to be installed on the cluster.

This also has a secondary benefit: smaller schemas (faster introspection, less client tooling weight).

## Proposed change

Add an optional schema-filter section to `ClusterAccess.spec` that selects which resource types are generated. Each entry selects by **group** (required), with **kind** and **version** as optional narrowing fields:

```yaml
apiVersion: gateway.platform-mesh.io/v1alpha1
kind: ClusterAccess
metadata:
  name: public-api
spec:
  host: https://<cluster-api-server>
  auth:
    serviceAccountRef:
      name: graphql-gateway
      namespace: graphql-gateway
  schemaFilter:
    include:
      # group only — every kind, every version in this group is generated
      - group: example.platform-mesh.io

      # group + kind — that kind across all its versions
      - group: example.platform-mesh.io
        kind: Widget

      # group + kind + version — a single, pinned GVK
      - group: example.platform-mesh.io
        kind: Gadget
        version: v1alpha1

      # group + version, no kind — every kind at that version of the group
      - group: ""
        version: v1
```

Semantics:
- **`group`** is required.
- **`kind`** is optional — omit it to include every kind in the group.
- **`version`** is optional — omit it to include all served versions of the matched group/kind.

An entry matches the most specific of these that is set; anything not matched by any `include` entry is omitted from the schema entirely.

Open questions for the maintainers (happy to align on any of these):
- **Allowlist vs. denylist:** an `include` allowlist covers our use case; a matching `exclude` denylist could be useful for others. Allowlist alone would be enough for us.
- **Where the filter is applied:** conceptually this fits as a filtering step in the listener's schema-resolution pipeline (after the cluster's types are loaded, before the schema is marshalled/streamed), so filtered-out types never reach the gateway at all. Applying it there — rather than at query time in the gateway — keeps the generated schema itself minimal, which is the property we care about.
- **Config location:** on `ClusterAccess.spec` (per-cluster) is the most useful for us, since the filter is a property of *this particular exposed API*. A listener-wide flag would be less flexible.

## Alternatives considered

- **Filtering downstream at the consumer** — doesn't help; the internal types are still present in the schema and reachable, so it's neither a real boundary nor a smaller schema.
- **A dedicated cluster with only the desired CRDs installed** — works but is operationally expensive and duplicates infrastructure just to shape an API surface.
- **A listener deployment argument (flag/env) instead of a field on `ClusterAccess`** — simpler to implement and fine when a listener serves a single, uniform API surface. The downside is that it applies to every cluster that listener handles, so it can't express different filters for different `ClusterAccess` resources. We'd be happy with this if per-cluster configuration is out of scope, but the per-`ClusterAccess` field is more flexible.

## Notes

Understood that the project is in an alpha stage and APIs may change — treat the concrete field names/shape above as a starting point for discussion, not a fixed proposal. Happy to help refine the design or contribute the implementation if that's useful.

hellsontime:
## Summary

Adds an optional `schemaFilter` to `ClusterAccess`, allowing each GraphQL endpoint to expose typed operations only for selected Kubernetes API groups, versions, and kinds.

If no filter is configured, the existing behavior is preserved and all discovered resources remain available.

## Design decision

This PR treats schema filtering as GraphQL schema shaping, not as an authorization boundary. Kubernetes RBAC remains authoritative, and generic operations such as `applyYaml` are intentionally unchanged.

This keeps the initial implementation focused, but we are open to adjusting the behavior if full enforcement across generic operations is expected.

Closes #197.

## Testing

- Added unit tests for selector matching, reference closure, resolver integration, and GraphQL schema generation.
- Added controller/envtest coverage for filtered and unfiltered `ClusterAccess` objects.
- Ran generation, linting, unit, e2e, and envtest suites successfully.
- Manually verified against a real kind cluster:
  - Pods and Deployments remained available through the filtered endpoint.
  - ConfigMaps were excluded from introspection and rejected by GraphQL validation.
  - The same ConfigMaps query succeeded through an unfiltered control endpoint.


apis/gateway/v1alpha1/clusteraccess_types.go
```diff
@@ -63,6 +63,32 @@ type ClusterAccessSpec struct {
...
+	// Group is the Kubernetes API group. Use an empty string for the core API group.
+	Group string `json:"group"`
+
+	// Version is the optional Kubernetes API version to match.
+	// +optional
+	Version string `json:"version,omitempty"`
+
+	// Kind is the optional Kubernetes resource kind to match.
+	// +optional
+	Kind string `json:"kind,omitempty"`
```
xrstf: Last chance to prevent a mistake I made in the syncagent: let's not conflate resources (lowercase, pluralized) with kinds (uppercase, singular). If this is a ResourceSelector, it should select a _resource_ (i.e "pods" instead of "Pod").

services/kubernetes-graphql-gateway/apischema/schemaset_filter.go
```diff
@@ -0,0 +1,125 @@
...
+// ResourceSelector matches a Kubernetes resource by exact group and optional
+// exact version and kind.
+type ResourceSelector struct {
+	Group   string
+	Version string
+	Kind    string
+}
+
+// Matches reports whether the selector matches the supplied GVK.
+func (s ResourceSelector) Matches(gvk schema.GroupVersionKind) bool {
```
xrstf: Yeah we should really settle on whether this is all about Kinds or all about resources.

services/kubernetes-graphql-gateway/apischema/schemaset_filter.go
```diff
@@ -0,0 +1,125 @@
...
+	for _, selector := range selectors {
+		if selector.Matches(gvk) {
+			return true
+		}
+	}
+
+	return false
+}
+
+func schemaReferences(schema *spec.Schema) map[string]struct{} {
```
xrstf: We already import apimachinery, that module has a nice `Set` datatype that we can use here instead of a custom string-struct map.

apis/gateway/v1alpha1/clusteraccess_types.go
hellsontime: agreed, since this is selecting the exposed API surface, resources are the better fit, changed it accordingly.

PR #302, opened 2026-08-25
fix(search-service): return 401 for invalid JWTs

### LINKED ISSUE #256: task: Return 401 for invalid JWT on all API requests
### Overview

# Current Behavior

The search service does not validate JWT tokens. It only parses them without signature verification (UnsafeClaimsWithoutVerification), delegating that to the service mesh. When a token is structurally valid but logically invalid (expired, wrong issuer, wrong audience), it still parses successfully. The user string is extracted from claims and passed to OpenFGA, which denies all permission checks. The result: the caller gets 200 OK with an empty result set instead of a 401 Unauthorized.

# Desired Behavior

Every API request must validate the JWT before processing. An invalid token must return 401 Unauthorized with no results.

# Acceptance Criteria

  - Requests with a missing or structurally unparseable token return 401
  - Requests with a token that fails standard JWT claims validation (expired, wrong issuer, wrong audience) return 401
  - Valid tokens continue to work as before (org access check → FGA → results)
  - The validation runs in middleware, before any business logic or FGA calls are made

### Implementation Details

_No response_

### Completion Checklist

- [ ] Documentation updated
- [ ] Changes reviewed

pteich:
The search-service now returns `401 Unauthorized` when JWT validation fails, including expired tokens and tokens with an invalid issuer or audience.

JWT validation happens in middleware before search business logic or OpenFGA calls. Missing and structurally invalid tokens also continue to return 401.

Local development behavior is preserved: `--is-local=true` uses the configured local organization and bypasses kcp JWT validation.

Fixes https://github.com/platform-mesh/backlog/issues/256

### DIFF
--- a/services/search-service/internal/clients/kcp/access_validator.go
+++ b/services/search-service/internal/clients/kcp/access_validator.go
@@ -31,7 +31,7 @@ import (
 
 const (
 	// maxSnippetBytes defines how much of an unexpected response body to read for logging.
-	maxSnippetBytes = 1 << 10
+	maxSnippetBytes = 1024
 )
 
 type OrgAccessValidator struct {
--- a/services/search-service/internal/clients/kcp/access_validator.go
+++ b/services/search-service/internal/clients/kcp/access_validator.go
@@ -29,6 +29,11 @@ import (
 	"k8s.io/client-go/rest"
 )
 
+const (
+	// maxSnippetBytes defines how much of an unexpected response body to read for logging.
+	maxSnippetBytes = 1 << 10
+)
+
 type OrgAccessValidator struct {
 	http    *http.Client
 	baseURL *url.URL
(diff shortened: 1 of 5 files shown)


services/search-service/internal/clients/kcp/access_validator.go
```diff
@@ -29,6 +29,11 @@ import (
 	"k8s.io/client-go/rest"
 )
 
+const (
+	// maxSnippetBytes defines how much of an unexpected response body to read for logging.
+	maxSnippetBytes = 1 << 10
```
FWuermse: Why would this be more than one byte? Should we not be able to detect an unexpected response from the first byte?

FWuermse (dismissed):

services/search-service/internal/clients/kcp/access_validator.go
xrstf: It's not about detecting, but logging the unexpected payload. Personally I would not use this syntax, but instead just write 1024 since it's much, much, much easier to grasp.

pteich: yes, it's to get payload for logging. I was used to the bit shift syntax for expressing bytes, because it's quite handy to express higher numbers. But I agree, in this case 1024 is way easier - changed it.

xrstf (approved):

PR #303, opened 2026-08-25
fix(marketplace): preserve permission claim selectors

anastazya:
Marketplace currently projects a legacy v1alpha1 APIExport and does not expose
the authoritative v1alpha2 permission-claim verbs and selectors. A UI cannot
accept a scoped claim without either losing information or broadening it.

This change keeps the existing v1alpha1 `spec.apiExport` projection and adds
`spec.apiExportPermissionClaims` with the exact v1alpha2 claims. The storage
projection:

- preserves verbs, `defaultSelector`, labels, and match expressions;
- exposes only wildcard claims through the legacy projection;
- omits v1alpha1 name/namespace `resourceSelector` claims that cannot be
  represented safely in v1alpha2; and
- fails closed on duplicate group/resource/identity keys.

The generated CRD, APIResourceSchema, APIExport, and deep-copy code are updated
from the API types. The APIResourceSchema uses a new immutable revision name.

Validation:

- `go test -count=1 ./...` in `apis`;
- `go test -count=1 ./...` in `services/virtual-workspaces`;
- 10-second Marketplace round-trip fuzz run;
- generated resource/APIExport consistency tests; and
- `git diff --check`.

The matching Marketplace UI change must be deployed after this server change.
This additive contract is intended to supersede the breaking nested-type
replacement in platform-mesh/platform-mesh#274.


anastazya:
Closing in favor of #274 plus the end-to-end UI migration discussed on Zulip — the alpha2 export already serves the exact claims. The fail-closed binding composition moves to a follow-up on top of marketplace-ui#187.

PR #317, opened 2026-08-27
Fix: Use User-ID instead of e-mail

j-holub:
We want to use user ids instead of the e-mail to identify a user and query for their workspaces and other entities. This makes the necessary changes.

### DIFF
--- a/operators/security-operator/internal/subroutine/invite.go
+++ b/operators/security-operator/internal/subroutine/invite.go
@@ -18,18 +18,13 @@ package subroutine
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
-	"net/http"
 
-	"github.com/coreos/go-oidc"
 	"github.com/rs/zerolog/log"
-	"golang.org/x/oauth2/clientcredentials"
 
 	pmcorev1alpha1 "go.platform-mesh.io/apis/core/v1alpha1"
 	"go.platform-mesh.io/golang-commons/controller/lifecycle/ratelimiter"
 	iclient "go.platform-mesh.io/security-operator/internal/client"
-	"go.platform-mesh.io/security-operator/internal/config"
 	"go.platform-mesh.io/subroutines"
 
 	"k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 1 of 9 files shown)


operators/security-operator/internal/subroutine/invite.go
```diff
@@ -141,3 +181,38 @@ func (w *inviteSubroutine) reconcile(ctx context.Context, obj ctrlruntimeclient.
...
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
+	if err != nil {
+		return "", fmt.Errorf("creating request: %w", err)
+	}
+
+	res, err := w.keycloak.Do(req)
+	if err != nil {
+		return "", fmt.Errorf("querying Keycloak user: %w", err)
+	}
+	defer res.Body.Close() //nolint:errcheck
```
xrstf: Since we want to use / make possible other IdPs than Keycloak, do we know if

a) this affects non-Keycloak Platform-Meshes (those cannot exist right now, but should hopefully soon-ish-ish), and
b) whether other IdPs generally have a similar API function to mimic Keycloak?

Because if we're adding new code here that can only ever work with Keycloak, and is in a code path that affects every Platform Mesh, this PR might bring us into trouble.

j-holub: Thank you very much for the thorough review. This part is outdated, we decided that it's best to keep the invite to e-mail and I reverted this part of the PR.

xrstf:
Is this a fix or a change or even a breaking change? I couldn't really judge and would have to defer to the @platform-mesh/tsc for opinion.

services/search-service/internal/config/config.go
```diff
@@ -24,7 +24,7 @@ import (
 	"github.com/spf13/pflag"
 )
 
-const defaultUserClaim = "email"
```
FWuermse, line 27: For now default must stay `email` to comply with the default settings of the security operator.

j-holub: Yeah makes sense. Let's close this

j-holub:
We don't want to change the defaults. This change is not needed.

PR #322, opened 2026-08-28
security-operator: generate authorization models for self-bound APIExports

xmudrii:
## Problem

Composite types published by KROaaS get no OpenFGA type, so the portal denies listing their instances. The RGD reports Active/Ready, so nothing surfaces the failure.

## Cause

`Process` resolves ProviderPermissions via `GetCluster(ProvidersProviderName, ...)` and returns on error. Only provider workspaces are in providers scope, and KROaaS self-binds its per-RGD APIExport in the consumer's own workspace, so it aborts before generating a model. Added by 511d6688, first shipped in v0.35.0.

## Fix

Skip the lookup when the export is self-bound (`APIExportClusterName` equals the binding's own cluster). Those exports have no ProviderPermissions to read and fall back to the template defaults, as v0.34.3 did unconditionally. Every other export keeps the v0.35.0 path.

Merely making the lookup non-fatal would be fail-open: `DefaultPermissions` can be stricter than the defaults, so a transient failure would overwrite a provider's restrictive model with a permissive one.

## Testing

Three cases added to `TestAuthorizationModelGeneration_Process`; the self-bound one fails without the fix. Unit and integration suites pass. Verified on local-setup.

### DIFF
--- a/operators/security-operator/internal/subroutine/authorization_model_generation.go
+++ b/operators/security-operator/internal/subroutine/authorization_model_generation.go
@@ -272,7 +272,6 @@ func (a *AuthorizationModelGenerationSubroutine) Process(ctx context.Context, ob
 	internalAPIBindings := []string{"core.platform-mesh.io", "system.platform-mesh.io", "providers.platform-mesh.io"}
 
 	if slices.Contains(internalAPIBindings, binding.Spec.Reference.Export.Name) || strings.HasSuffix(binding.Spec.Reference.Export.Name, "kcp.io") {
-		// If the APIExport is the core.platform-mesh.io, system.platform-mesh.io, providers.platform-mesh.io we can skip the model generation.
 		return subroutines.OK(), nil
 	}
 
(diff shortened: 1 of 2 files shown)


operators/security-operator/internal/subroutine/authorization_model_generation.go
xrstf: While we're here, let's remove this comment. It adds nothing and is also outdated since it ignores the `or` condition.

operators/security-operator/internal/subroutine/authorization_model_generation.go
```diff
@@ -298,22 +298,31 @@ func (a *AuthorizationModelGenerationSubroutine) Process(ctx context.Context, ob
...
+	var providerPermissions *pmprovidersv1alpha1.ProviderPermissions
 
-	var providerPermissionsList pmprovidersv1alpha1.ProviderPermissionsList
-	err = providerCluster.GetClient().List(ctx, &providerPermissionsList)
-	if err != nil {
-		return subroutines.OK(), fmt.Errorf("listing ProviderPermissions: %w", err)
-	}
+	// A self-bound export has no provider workspace and uses the template defaults. Every
+	// other export must resolve, or the model would lose the provider's permission overrides.
+	selfBound := binding.Status.APIExportClusterName != "" &&
```
xrstf: The `binding.Status.APIExportClusterName != ""` check is redundant, no? The `logicalcluster.From(binding)` can never be empty, can it?

operators/security-operator/internal/subroutine/authorization_model_generation.go
xmudrii: Removed.

operators/security-operator/internal/subroutine/authorization_model_generation.go
xmudrii: Removed.

xrstf (approved):

PR #332, opened 2026-08-31
fix(marketplace): use identityHash to skip unestablished APIExports

lpgarzonr:
## What

In `services/virtual-workspaces/pkg/storage/filter.go`, the `Marketplace()` function skips `APIExport`s with `len(latestResourceSchemas) == 0`. This PR replaces that condition with `export.Status.IdentityHash == ""`.

## Why

The `latestResourceSchemas` check was intended to skip exports that are not yet established. But it conflates two distinct states:

1. **Not yet established** — kcp has not set the identity hash; the export cannot accept bindings yet.
2. **UI-only** — the provider intentionally exposes no CRD (no resource schemas), but is fully established and ready to accept bindings.

`status.identityHash` is the correct readiness signal. Looking at the kcp APIExport reconciler (`pkg/reconciler/apis/apiexport/apiexport_reconcile.go`), the lifecycle is:
- **Reconciliation 1**: kcp creates the identity secret and sets `spec.identity.secretRef`, then returns early — `status.identityHash` is still empty at this point.
- **Reconciliation 2**: kcp reads the secret, SHA-256 hashes the key, and writes the result to `status.identityHash`. Once set it is immutable.

`identityHash` is therefore the authoritative signal that the export is established and can accept bindings — independent of whether any `APIResourceSchema`s are referenced.

## Impact

With the previous check, any UI-only provider (portal navigation / micro-frontend, no operator CRD) is permanently excluded from the Marketplace catalog even after its workspace resources are fully applied. All four existing local-setup providers happen to have schemas so this was never hit — but migrating the first UI-only extension (`cloud-orchestrator`) uncovered the bug.

After this change, UI-only providers appear in the Marketplace as soon as `identityHash` is set.

## Side effects and edge cases

**Transient empty hash (first reconciliation window):** An `APIExport` will briefly have an empty `identityHash` between the first and second reconciliation cycle (typically sub-second). During this window the Marketplace skips it — identical to the previous behavior where a schema-less export was always skipped. This self-resolves.

**Empty `APIBinding` for UI-only exports:** When a UI-only export is installed, the `APIBinding` is created with an empty bound API group — there are no schemas for kcp to project. This is intentional: the binding scopes the consumer workspace to the provider, and the portal renders the `ContentConfiguration`. No functional regression for existing schema-carrying providers.

**Exports with schemas:** Behavior is unchanged — they have a non-empty `identityHash` once established, so they continue to appear exactly as before.

## Changes

- `filter.go`: one-line guard change.
- `marketplace_test.go`: updated `makeExport` fixture to set `Status.IdentityHash` (required by the new check); added two new tests:
  - `TestMarketplace_UIOnlyExportWithNoSchemas` — export with no schemas but established `identityHash` appears in Marketplace.
  - `TestMarketplace_UnestablishedExportSkipped` — export with schemas but empty `identityHash` is still skipped.

Discovered while validating the HSP → Platform Mesh extension migration end-to-end on a local kind cluster (DXPFRAME-2475).

### DIFF
--- a/services/virtual-workspaces/pkg/storage/filter.go
+++ b/services/virtual-workspaces/pkg/storage/filter.go
@@ -241,7 +241,7 @@ func Marketplace(
 					if logicalcluster.From(&export) != providerClusterID {
 						continue
 					}
-					if len(export.Spec.LatestResourceSchemas) == 0 {
+					if export.Status.IdentityHash == "" {
 						continue
 					}
 
--- a/services/virtual-workspaces/pkg/storage/marketplace_test.go
+++ b/services/virtual-workspaces/pkg/storage/marketplace_test.go
@@ -84,6 +84,9 @@ func makeExport(cfg config.ServiceConfig, name, clusterID, provider string) *kcp
 		Spec: kcpapisv1alpha1.APIExportSpec{
 			LatestResourceSchemas: []string{"v260810" + name},
 		},
+		Status: kcpapisv1alpha1.APIExportStatus{
+			IdentityHash: "testhash-" + name,
+		},
 	}
 }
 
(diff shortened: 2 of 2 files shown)


services/virtual-workspaces/pkg/storage/filter.go
```diff
@@ -241,7 +241,7 @@ func Marketplace(
 					if logicalcluster.From(&export) != providerClusterID {
 						continue
 					}
-					if len(export.Spec.LatestResourceSchemas) == 0 {
+					if export.Status.IdentityHash == "" {
```
brsmnv, line 244: This looked a bit weird but it is apparently correct! 

The relevant  part from the kcp side: https://github.com/kcp-dev/kcp/blob/v0.32.3/pkg/reconciler/apis/apiexport/apiexport_reconcile.go#L92-L102

Identity hash is immutable and apparently it can go stale. So the if "" { continue} part is correct, but we should add one check. Right after this block:
```go
import(
	"github.com/kcp-dev/sdk/apis/third_party/conditions/util/conditions"
	kcpapisv1alpha1 "github.com/kcp-dev/sdk/apis/apis/v1alpha1"
// ...


  if export.Status.IdentityHash == "" { continue}
  // check condition:
  if conditions.IsFalse(&export, kcpapisv1alpha1.APIExportIdentityValid) {
    klog.V(2).InfoS("... APIExport with invalid identity", args...)
  }

```


A once valid export can get out of sync wrt the identity secret in the provider workspace. And once the hash no longer matches, the condition will flip to false: https://github.com/kcp-dev/kcp/blob/08be0a1c8de29f5ec6fc27d3563288be970bd3b4/pkg/reconciler/apis/apiexport/apiexport_reconcile.go#L207-L209

brsmnv: Honestly, I would actually skip the log line and the check. Not sure what we would be supposed to do with that in the logs.

lpgarzonr: Thanks for digging into the kcp side! Good point on the stale identity case — agreed that adding a log line without a clear recovery action would just be noise. Happy to leave the fix as-is.

brsmnv: Resolved,  nothing to handle here :)

aaronschweig (approved):

PR #333, opened 2026-09-01
split core.platform-mesh.io into separate APIExports

### LINKED ISSUE #47: Hide Store  resource from user workspaces
## Context

The `Store` and `AuthorizationModel` resources are currently exposed via the `core.platform-mesh.io` APIExport, making them visible in every user workspace. These are internal technical resources that end users should not modify directly.

## Goal

Restrict visibility of these resources to organizational workspaces only:

1. **Create a new APIExport** `orgs.core.platform-mesh.io` containing `Store` and `AuthorizationModel` resources
2. **Add this APIExport** to the default API bindings for organizational workspaces only
3. **Relocate AuthorizationModel generation** from provider workspaces to organizational workspaces
   - Ensure naming conventions prevent conflicts across workspaces
4. **Update affected components:**
   - Adjust generator/operator logic accordingly
   - Update helm-charts default configurations

## Acceptance Criteria

- [ ] `Store` and `AuthorizationModel` resources are no longer visible in user workspaces
- [ ] New `orgs.core.platform-mesh.io` APIExport is created and bound to org workspaces
- [ ] AuthorizationModels are generated in org workspaces with conflict-safe naming
- [ ] All related operators and helm-charts are updated

### LINKED ISSUE #171: Split monolithic core.platform-mesh.io APIExport into self-contained per-component APIExports
# Problem
The core.platform-mesh.io APIExport bundles latestResourceSchemas from multiple, unrelated API groups:

`core.platform-mesh.io` — Account, AccountInfo, AuthorizationModel, Store, Invite (account-operator / L1)
`ui.platform-mesh.io` — ContentConfiguration, ProviderMetadata (portal / marketplace / L4)

Because they share one export, a workspace can't consume one group without binding all of them. 
Concretely: a provider workspace needs the `ui.platform-mesh.io` kinds (provider bootstrap writes `ContentConfiguration/ProviderMetadata` if UI is enabled), but the only way to get them today is to bind core, which also drags in `Account/Invite/Store` and ties an L4 concern to an L1 export. In the modular monorepo this also blocks moving the providers controller, since it can't depend on a self-contained UI API.

**Security concern: Because currently this drags in rest of the APIs, provider, managing its own provider code can interact with these object and create chaos in the system by creating accounts, invites, etc.** 

# Goal

Each API group is exported by its own component, as a self-contained APIExport, so consumers bind exactly what they need and layer boundaries (RFC 008) hold.

# Proposed change

* Introduce a dedicated `ui.platform-mesh.io` APIExport (schemas: `ContentConfiguration`, `ProviderMetadata`), **(TBC Maybe even separate ProviderMetadata and `ContentConfiguration` I dont think they are both part of UI???)**
owned by the `UI/portal/extension-manager` component; remove those two schemas from `core.platform-mesh.io`.

* Keep `core.platform-mesh.io` as account-operator's own export (`Account/AccountInfo/AuthorizationModel/Store/Invite` only), owned by account-operator. **(MAYBE WE create new one, and deprecate core all togheter?)**

* Update defaultAPIBindings on the relevant `WorkspaceTypes` to bind the new self-contained exports (e.g. the provider `WorkspaceType` does not have `ui`, nor core. So UI is separate binding request).

# Acceptance criteria
* ` ui.platform-mesh.io` served via its own `APIExport`, independent of core.
 * `core.platform-mesh.io` exports only the account/core kinds.
 * A provider workspace can create `ContentConfiguration/ProviderMetadata` without binding core when UI is enabled in the provider workspace (not part of WST!, separate flow!)
 * No cross-group schemas remain in any single APIExport.
* Existing consumers (accounts, portal) still resolve after the split (bindings updated).

# Notes
* Unblocks the providers-controller migration into the monorepo (currently deferred on this coupling).
Schema/identity-hash changes to core mean existing APIBindings must be re-pointed — call out migration/ordering for live workspaces.

iakmc:
## Summary

- Splits the `core.platform-mesh.io` APIExport into three: `core.platform-mesh.io`, `ui.platform-mesh.io`, `orgs.core.platform-mesh.io`.
- `core.platform-mesh.io` now only has `accountinfos`, `accounts`, `invites`.
- `ui.platform-mesh.io` (new): `contentconfigurations`, `providermetadatas`.
- `orgs.core.platform-mesh.io` (new): `authorizationmodels`, `stores`. Bound only on org workspaces (`workspace-type-orgs.yaml`), not on every workspace type.
- `security-operator` updated too, it now watches Store and AuthorizationModel through `orgs.core.platform-mesh.io`, not `core.platform-mesh.io` anymore.
- `platform-mesh-operator` now migrates existing `root:orgs` workspaces off the old binding automatically, no manual step needed anymore.

## What fixes

Before this, any workspace that bound `core.platform-mesh.io` to reach one kind got all of them. A provider workspace binding core just to get `ContentConfiguration`/`ProviderMetadata` also got write access to `Account`/`Invite`/`Store`, a real security boundary issue, not just messy grouping.

- Closes #171
- Closes #47

## Testing

- `go build ./...` passes, both operators.
- `task verify-security-operator` and `task verify-platform-mesh-operator` both pass clean (lint, tidy-check, boilerplate).
- `task test-security-operator` and `task test-platform-mesh-operator` both pass, all packages.
- Full e2e suite passes, all subtests, including full `PlatformMesh` readiness.
- Tested the real upgrade path on one cluster: platform mesh from the code before this change, installed a real provider, confirmed working. Then applied this change plus the security-operator fix on the same cluster. Old provider kept working, zero restarts. New provider installed after also worked, got UI access automatically through the normal opt-in (`extraDefaultAPIBindings`).
- Manually verified against live kcp:
  - `core.platform-mesh.io` no longer serves `stores`/`authorizationmodels`/`contentconfigurations`.
  - `ui.platform-mesh.io` is `Bound`, existing `ContentConfiguration` objects still readable.
  - `orgs.core.platform-mesh.io` auto-bound on `root:orgs`, existing `Store` object still readable.
- Two real bugs found and fixed during testing, both proven with a genuine fresh-cluster before/after upgrade, not just reasoning:
  - security-operator's watcher for `orgs.core.platform-mesh.io` was scoped wrong and never matched any cluster at all, so `Store`/`AuthorizationModel` never reconciled through it. Fixed the scoping.
  - Existing `root:orgs` workspaces never moved off the old binding on their own. `platform-mesh-operator` now does this automatically every reconcile, clearing finalizers safely so the old binding can actually finish releasing them, no manual step needed.
- Not covered here: relocating `AuthorizationModel` generation from provider to org workspaces, issue #47 Goal 3, that's a separate change in `security-operator`.


operators/platform-mesh-operator/manifests/kcp/01-platform-mesh-system/apiexport-core.platform-mesh.io.yaml
```diff
@@ -29,26 +29,6 @@ spec:
     schema: v260109-82344be.accounts.core.platform-mesh.io
     storage:
       crd: {}
-  - group: ui.platform-mesh.io
```
mjudeikis, line 32: :root:providers:foo - needs UI apis currently. How it will get them?

operators/platform-mesh-operator/manifests/kcp/02-root/01-workspace-type-orgs.yaml
```diff
@@ -22,6 +22,8 @@ spec:
       path: root:platform-mesh-system
     - export: system.platform-mesh.io
       path: root:platform-mesh-system
+    - export: orgs.core.platform-mesh.io
```
mjudeikis, line 25: is this breaking change if we apply this to PM instance before this change and  with existing orgs?

operators/platform-mesh-operator/manifests/kcp/01-platform-mesh-system/apiexport-core.platform-mesh.io.yaml
mjudeikis: Like how these apis lands in there https://github.com/platform-mesh/provider-quickstart/blob/main/config/provider/providermetadata.yaml ?

iakmc: I tested this many times now. Full scenario: fresh cluster on old code, install real provider (wildwest), then apply this PR and the security-operator fix on same cluster, then install second provider after.

The file you linked (config/provider/providermetadata.yaml) is embedded in the provider's own binary and applied by its init container to its own workspace, on start. I saw it fail exactly like this, error no matches for kind "ProviderMetadata", when the workspace has no ui.platform-mesh.io binding.

How it gets the binding: PlatformMesh.spec.kcp.extraDefaultAPIBindings, set by whoever runs the deployment, workspaceTypePath: root:provider, export: ui.platform-mesh.io. If this is set before the provider workspace gets created, it binds automatically, no manual step. I checked this directly with admin kubeconfig, not only from pod being healthy.

One thing to know, this only works for workspaces created after the setting exists, kcp gives default bindings only at creation time. A provider workspace that already existed before this change is fine either way, it keeps using whatever binding it already had.

operators/platform-mesh-operator/manifests/kcp/02-root/01-workspace-type-orgs.yaml
iakmc: I created platform mesh from scratch on old code, created org and account, installed real provider, confirmed working. Then applied this PR and the security-operator fix on the same running cluster. Checked the old provider again after, still fine, no restarts, its ContentConfiguration and ProviderMetadata still there, served through the old core.platform-mesh.io binding, which keeps working since kcp does not take away access just because export got smaller. So nothing breaks right away for what is already running.

But the real migration, moving Store and AuthorizationModel to actually use the new export in an existing workspace, does not happen alone. New binding gets stuck against old one, and removing old one also gets stuck, kcp will not finish removing binding while live objects still use it. Root cause, security-operator only watched Store and AuthorizationModel through old export, so it could not see them anymore to remove its own finalizers. Fixed this in security-operator, also checked it is safe to deploy before this export split even lands.

After that I installed second, new provider on same cluster, next to the first one, and it worked too, ui.platform-mesh.io bound by itself, same way as above. So, old stuff keeps working, new stuff works correctly, only one manual step is still needed if someone wants to fully migrate a workspace that already has live Store/AuthorizationModel data. That part is not automatic, should be written down as migration step for whoever does real rollout.

operators/platform-mesh-operator/manifests/kcp/01-platform-mesh-system/apiexport-core.platform-mesh.io.yaml
mjudeikis: So what all this tells? This works or does not? :D

operators/platform-mesh-operator/manifests/kcp/01-platform-mesh-system/apiexport-orgs.core.platform-mesh.io.yaml
```diff
@@ -0,0 +1,31 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: apis.kcp.io/v1alpha2
+kind: APIExport
+metadata:
+  name: orgs.core.platform-mesh.io
```
mjudeikis, line 18: I have beef with this naming. Feels veryu much java style :D
All this is tenancy related thing, why can't we have `tenancy.platfrom-mesh.io` ?

Where these names are coming from?

operators/platform-mesh-operator/manifests/kcp/02-root/01-workspace-type-orgs.yaml
mjudeikis: > which keeps working since kcp does not take away access just because export got smaller. So nothing breaks right away for what is already running.

I doubt. If apiexport stopped serving/acting on the apis, how can this still work?

All second paragraph. What does it mean? Can you dummy-down it for me?

mjudeikis:
I will be bluntly honest - I dont understand if this PR work, what migration steps one needs todo (if any), or it does not? There are so much verbosity in the comments I just got lost :D

operators/platform-mesh-operator/manifests/kcp/02-root/01-workspace-type-orgs.yaml
iakmc: Checked kcp code directly on this one. When APIBinding connects to an APIExport, kcp makes a CRD for each resource and just keeps it. If a resource gets removed from the export list later, kcp does NOT delete that CRD, it just stops looking at it. So old workspace still has its CRD, keeps working, nothing breaks. kcp even has a comment about this exact thing.

Loop only adds, never removes: https://github.com/kcp-dev/kcp/blob/v0.32.3/pkg/reconciler/apis/apibinding/apibinding_reconcile.go#L219-L335
The TODO comment: https://github.com/kcp-dev/kcp/blob/v0.32.3/pkg/reconciler/apis/apibinding/apibinding_reconcile.go#L280

operators/platform-mesh-operator/manifests/kcp/01-platform-mesh-system/apiexport-orgs.core.platform-mesh.io.yaml
iakmc: Name is literally what the issue says, not me inventing java-style stuff. https://github.com/platform-mesh/platform-mesh/issues/47 says "Create a new APIExport orgs.core.platform-mesh.io containing Store and AuthorizationModel resources". I do not have anything against renaming this to tenancy.platform-mesh.io or whatever if you want, just took the name straight from the ticket.

operators/platform-mesh-operator/manifests/kcp/01-platform-mesh-system/apiexport-core.platform-mesh.io.yaml
iakmc: Yes, works. Old provider keep running same as before after upgrade, new provider get ui.platform-mesh.io automatic through extraDefaultAPIBindings, set before workspace creation. Existing workspaces from before the change just keep old binding, nothing breaks either way.

And MJ, if you asking what extraDefaultAPIBindings is :D, here short version. It's setting on our PlatformMesh CR. You give it workspace type (like "provider") + export (like "ui.platform-mesh.io"). This is not fixed in our code, person who installs platform mesh chooses it, can turn it on or off, depends on the environment. kcp already have own thing for auto give APIs to new workspaces, called defaultAPIBindings, our setting just add entry there before workspace type gets created. End result, new provider workspace get ui.platform-mesh.io automatic, zero manual steps. Only catch, works only for workspaces created after you set it, kcp don't go back and fix old ones.

Field: https://github.com/platform-mesh/platform-mesh/blob/main/apis/core/v1alpha1/platformmesh_types.go#L91
Applied here: https://github.com/platform-mesh/platform-mesh/blob/main/operators/platform-mesh-operator/pkg/subroutines/subroutine_helpers.go#L629-L651
Match logic: https://github.com/platform-mesh/platform-mesh/blob/main/operators/platform-mesh-operator/pkg/subroutines/kcpsetup.go#L426-L440

operators/platform-mesh-operator/manifests/kcp/02-root/01-workspace-type-orgs.yaml
mjudeikis: > ContentConfiguration and ProviderMetadata still there

object will be there, but is there anything actign on them? If apiexport got dropped these object, they will not be reflected in apiexport-virtual-workspace. so lets say you want to edit ContentConfiguration, will it be applied to the platfrom?

operators/platform-mesh-operator/manifests/kcp/01-platform-mesh-system/apiexport-orgs.core.platform-mesh.io.yaml
mjudeikis: I think this might need some small topic in zulip to discuss

iakmc:
> I will be bluntly honest - I dont understand if this PR work, what migration steps one needs todo (if any), or it does not? There are so much verbosity in the comments I just got lost :D

Short answer, does it work? Yes.

- Nothing breaks for what already running today, even if you do nothing after this PR.
- New workspaces created after this PR, fully automatic, zero manual steps.
- Old workspaces that already exist with Store/AuthorizationModel data, one manual step needed if you want them actually moved to new export.
- What that step is: old workspace already has APIBinding to core.platform-mesh.io holding Store/AuthorizationModel, kcp locks those resources to that binding and doesn't release lock just because we removed them from the export later. To actually switch, someone deletes that old core.platform-mesh.io APIBinding in the workspace, that releases the lock, new orgs.core.platform-mesh.io binding takes over.
- That delete used to hang forever, because kcp don't finish deleting a binding while live Store/AuthorizationModel objects still using it, and security-operator couldn't clean its own finalizers off those objects, it was only watching them through old export, so it can't see them no more. This is the bug what this PR's security-operator fix resolve, so delete now actually finishes instead of hanging.
- So, one delete of old APIBinding per existing workspace that has Store/AuthorizationModel data, and that's it. Without that step they just keep working on old export too, nothing breaks, just not "migrated"

operators/platform-mesh-operator/manifests/kcp/02-root/01-workspace-type-orgs.yaml
iakmc: Object stays there, it get applied to platform too, I checked it locally.

Local kind cluster started on the old code(before this PR), wildwest provider installed, then the changes from this PR applied on that same running cluster, then wildwest-second provider installed after, both still running now.

Called the real endpoint UI use for nav content, path is `/services/contentconfigurations/clusters/<workspace>/apis/ui.platform-mesh.io/v1alpha1/contentconfigurations`. Called it once per provider. Both times gave same content back. (service is `services/virtual-workspaces`)

Looks like service do not care which export is bound, it always read content direct from root:platform-mesh-system workspace. So splitting the APIExports does not touch this at all.

PR #335, opened 2026-09-01
feat(kubernetes-graphql-gateway): add trusted ServiceAccount request mode

### LINKED ISSUE #304: feat(kubernetes-graphql-gateway): trusted mode — run gateway with a ServiceAccount identity for tokenless requests
## Is your feature request related to a problem? Please describe.

Today the gateway requires an end-user Bearer JWT on every non-discovery request and returns 401 when none is present. In fully trusted deployments — where the caller is a technical/system component and authentication & authorization are enforced **outside** the gateway (e.g. mTLS at the ingress or service mesh) — there is no end-user token. The gateway cannot be used in these scenarios even though access is already authorized upstream.

## Describe the solution you'd like

A configurable **trusted / ServiceAccount mode** in which the gateway:

- Accepts requests that carry **no user JWT** (no 401 on missing `Authorization`).
- Performs all Kubernetes access using **a configured ServiceAccount + RBAC** rather than a per-request user identity.
- Keeps the requirement that the SA's RBAC scopes what the gateway can do, so access control is still enforced — just at the SA level rather than per-user.

One possible direction: the ServiceAccount could be declared on (or co-located with) the `ClusterAccess` resource, so the identity is declared alongside the cluster it grants access to. A simpler alternative is to use the gateway's own pod ServiceAccount directly (see Alternatives). The exact design is left to maintainers.

## Describe alternatives you've considered

Two candidate sources for the ServiceAccount identity (not mutually exclusive):

1. An SA declared **on the `ClusterAccess` resource** — per-cluster declarative identity, co-located with the cluster-access configuration.
2. The **gateway's own pod ServiceAccount** used directly, with RBAC bound to it — simpler; one identity for the gateway process.

## Additional context

- **Scope:** gateway request-path auth only. External trust establishment (mTLS verification, ingress/mesh config) is explicitly out of scope and assumed handled by infrastructure.
- The gateway already serves **discovery** requests with its own credentials, so a tokenless-with-own-identity path is not architecturally foreign.
- Related issues for context: #253 (external federated technical-user identity — different mechanism), #233 / #234 (TokenReview auth behavior), #283 (SelfSubjectAccessReview abstraction).

hellsontime:
## Summary

Add an explicit per-`ClusterAccess` identity mode for trusted callers that do not have an end-user bearer token.

When `spec.requestIdentityMode` is set to `serviceAccount`, the gateway accepts tokenless requests, skips TokenReview, and executes all Kubernetes operations using the configured `auth.serviceAccountRef`. The existing user-token flow remains the default for backward compatibility.

## Configuration

```yaml
spec:
  requestIdentityMode: serviceAccount
  auth:
    serviceAccountRef:
      name: graphql-gateway
      namespace: graphql-gateway
```

Selecting `serviceAccount` requires `auth.serviceAccountRef`; this is enforced by CRD validation and checked again when endpoint metadata is loaded.

## Security considerations

Trusted endpoints must be protected by an external trust boundary such as mTLS at the ingress or service mesh. A caller-provided bearer token cannot replace or expand the configured ServiceAccount identity, and Kubernetes RBAC remains the authorization boundary. External trust establishment is outside the scope of this change.

## Testing

- Added unit, HTTP, CRD/controller, and e2e coverage, including a test proving that a privileged caller token cannot expand the configured ServiceAccount's RBAC.
- Ran the complete gateway and shared API validation suites and manually verified both identity modes against an isolated kind cluster.

Closes #304


services/kubernetes-graphql-gateway/gateway/gateway/endpoint/endpoint.go
```diff
@@ -143,24 +147,26 @@ func New(
...
+			if err != nil {
+				http.Error(w, "Service Unavailable", http.StatusServiceUnavailable)
+				return
+			}
+			if !authenticated {
+				// Verdict details (TokenReview status.error, cache-hit vs fresh)
+				// are logged by the validator.
+				log.FromContext(r.Context()).V(1).Info("request rejected: token failed TokenReview", "cluster", name)
+				http.Error(w, "Unauthorized", http.StatusUnauthorized)
+				return
```
aaronschweig: This nested if statement could be inverted with an early return to reduce nesting

aaronschweig (commented):
Nice work, thanks!

A few comments:

1. There is a slight inconsistency in naming:
- `usesServiceAccountForRequests` (endpoint.go:186)
- `Cluster.UsesServiceAccountForRequests()` (cluster.go)
- `Endpoint.AllowsTokenlessRequests()` / `Registry.AllowsTokenlessRequests()` / `Service.AllowsTokenlessRequests()`

It also seems like we are passing that information through every layer. Can we just verify that the information is needed everywhere or if there is potential to maybe trim it down a bit

services/kubernetes-graphql-gateway/gateway/http/http.go
```diff
@@ -87,29 +86,21 @@ func NewServer(c ServerConfig) (*Server, error) {
 
 		clusterName := r.PathValue("clusterName")
 
-		// Allow unauthenticated GET requests through to the playground handler.
-		if c.PlaygroundEnabled && r.Method == http.MethodGet {
-			ctx := utilscontext.SetCluster(r.Context(), clusterName)
-			queryHandler.ServeHTTP(w, r.WithContext(ctx))
-			return
-		}
-
```
aaronschweig, line 96: We still need the playground bypass.

Was this removed by accident or intentionally?

hellsontime: It was removed intentionally because a GET without an Authorization header now reaches the endpoint, where the existing playground bypass still serves it. The only behavior that changed is a playground GET with an explicitly malformed or empty Authorization header: it is now rejected by the HTTP layer instead of being ignored. Is bypassing header-format validation for playground requests the behavior you want to preserve?

aaronschweig: ah okay, thanks for the clarification

aaronschweig (approved):

PR #344, opened 2026-09-03
feat: improve exposure configuration in platform-mesh-operator

akafazov:
Enhance `spec.exposure` configuration in PlatformMesh CRD and use it in the template replacement code. This enables users to configure exposure once and propagate the value via templateing, thus reducing configuration.

refers to https://github.com/platform-mesh/helm-charts/issues/2458


OlegErshov (approved):
LGTM

PR #346, opened 2026-09-04
refactor(search-operator): remove hardcoded OpenSearch credentials

akafazov:
## What

Align the search-operator with platform-mesh standards for operator configuration and startup.

- Introduce a `cmd/` package (`root.go` + `operator.go`) with a cobra command tree
- Replace stdlib `flag` + `envconfig` with `cobra`/`pflag` + `golang-commons/config.CommonServiceConfig`
- Replace envconfig-tagged nested `Config` struct with a flat `OperatorConfig` using `AddFlags`
- Remove hardcoded default values for OpenSearch connection parameters
- OpenSearch credentials (`URL`, `Username`, `Password`) remain env vars — they are secrets
- Replace `NewClientFromEnv` (which redundantly called envconfig and re-read env vars) with `NewConfigFromEnv` + `NewClientFromConfig`
- Remove the `vrischmann/envconfig` dependency

## Why

Every other operator in the platform uses the golang-commons cobra/pflag pattern for configuration. The search-operator was using a bespoke stdlib `flag` + envconfig approach that diverged from the standard, and had hardcoded credential defaults (`admin`/`admin`) that are unsuitable for production.


operators/search-operator/internal/opensearch/client.go
```diff
@@ -79,36 +78,24 @@ func NewClient(cfg Config) (*Client, error) {
...
 	}
-
-	return NewClient(Config{
+	return Config{
 		URL:                url,
-		Username:           username,
-		Password:           password,
-		InsecureSkipVerify: insecure,
-	})
+		Username:           os.Getenv("OPENSEARCH_USERNAME"),
```
FWuermse: Aaron recently mentioned we want to move away from envs and towards CLI args. Should we also change it here?

operators/search-operator/internal/subroutine/apibinding_watcher.go
```diff
@@ -56,12 +56,17 @@ type apiBindingWatcherSubroutine struct {
...
 // searchConfigClient must be scoped to the provider workspace.
 // localCfg must be the admin kcp REST config.
-func NewAPIBindingWatcherSubroutine(mgr mcmanager.Manager, orgsClient ctrlruntimeclient.Client, localCfg *rest.Config, indexPrefix string, cfg config.Config) (lifecyclesubroutine.Subroutine, error) {
+func NewAPIBindingWatcherSubroutine(mgr mcmanager.Manager, orgsClient ctrlruntimeclient.Client, localCfg *rest.Config, indexPrefix string, cfg config.OperatorConfig) (lifecyclesubroutine.Subroutine, error) {
 	rootCfg, err := stripPathFromConfig(localCfg)
 	if err != nil {
 		return nil, err
 	}
 
+	providerByGroup := make(map[string]string, len(cfg.SearchableResources))
```
FWuermse: Is this still used? I think this is dead code.

operators/search-operator/main.go
```diff
@@ -17,224 +17,9 @@ limitations under the License.
...
-	var metricsAddr string
-	var enableLeaderElection bool
-	var probeAddr string
-	var kcpKubeconfig string
-	var apiExportEndpointSliceName string
-	var enableHTTP2 bool
-	var maxConcurrentReconciles int
-	var tlsOpts []func(*tls.Config)
-
-	flag.StringVar(&metricsAddr, "metrics-bind-address", "0", "The address the metrics endpoint binds to. "+
```
FWuermse, line 84: Where did the CLI args go?

FWuermse (changes requested):
Hey, thank you for the restructure! I just have some questions.

operators/search-operator/main.go
OlegErshov: Hi, they have been moved into the /internal/config/config.go file

operators/search-operator/internal/subroutine/apibinding_watcher.go
akafazov: yes, good catch!

operators/search-operator/internal/opensearch/client.go
akafazov: switched to cli args for those params

FWuermse (approved):

PR #148, opened 2026-06-30
feat: basic upstream identity provider implementation

simontesar:
Summary
Implements declarative upstream OIDC identity provider federation for organization realms via IdentityProviderConfiguration. Organizations can configure corporate SSO (Entra, Okta, Dex, …) in Platform Mesh instead of manually in the Keycloak admin console.

Part of https://github.com/platform-mesh/backlog/issues/286.

What this PR does

1. CRD & API (apis + generated schemas)
Adds spec.upstreamIdentityProviders[] with UpstreamIdentityProvider and OIDCUpstreamConfig
Adds provider-agnostic emailDomainRouting for identity-first login (domains, autoRedirect, hideUntilDomainMatch)
Adds status.managedUpstreamIdentityProviders per alias (ready, message, organizationId, linkedEmailDomains, lastSyncTime)
Updates CRD and KCP APIResourceSchema in security-operator and account-operator

2. security-operator reconciliation
Creates/updates/deletes Keycloak OIDC identity brokers from spec
Resolves client credentials from Secret refs in root:orgs (secrets are not stored in the CR)
Supports OIDC discovery URL or manual issuer/authorization/token endpoints
When emailDomainRouting is set: enables Keycloak Organizations on the realm, ensures org for domains, links broker with kc.org.* settings
Cleans up broker config and Keycloak org when upstream entries or domains are removed

3. Validation & seeding
Validating webhook for upstream spec (alias, OIDC config, domain rules, discovery vs manual endpoints)
Optional initializer IdP seed (--idp-seed-config-file) for local/dev profiles
Keycloak AdminClient extensions: organizations API, identity provider import/update, org–broker linkage

Important behavior
Upstream OIDC client must already exist on the corporate IdP; Platform Mesh configures Keycloak as that client — it does not register clients on the upstream side.
Keycloak broker redirect URIs must be allowlisted on the upstream IdP separately.
Portal Settings & Access UI is not in this PR (follow-up story under #308).

Test plan
 Manual: Dex upstream + identity-first login via local-setup (companion helm-charts PR)
 
Related
Epic: https://github.com/platform-mesh/backlog/issues/286
Design: https://github.com/platform-mesh/backlog/issues/286#issuecomment-4450331334
Local-setup / Dex / crds chart: separate helm-charts PR

PR #348, opened 2026-09-06
feat: IdPRegistration intent CR for org-owner upstream OIDC IdP

philtk79:
## Summary

Org owners cannot configure an upstream OIDC IdP today. The Portal UI has nothing to
call, because the only place upstream providers can live is
`IdentityProviderConfiguration` (IDPC) in `root:orgs` — a workspace tenants must not be
able to write.

This PR adds **`IdPRegistration`**, an intent CR that org owners create in their own
workspace (`root:orgs:<org>`). The security-operator watches it with an admin kubeconfig
and reconciles the Keycloak OIDC broker directly. Same pattern as `APIExportPolicy`: the
actor writes intent where they have rights, the operator does the privileged part
elsewhere.

Supersedes #148, which added upstream IdP support as a field on IDPC
(`spec.upstreamIdentityProviders`). That shape could not be exposed to tenants, since
IDPC is platform-owned and lives outside the org workspace.

Design: [`docs/design/upstream-identity-provider-registration.md`](docs/design/upstream-identity-provider-registration.md)
RFC context: [`contrib/rfc/rfc-01-delegated-idp-configuration.md`](contrib/rfc/rfc-01-delegated-idp-configuration.md)

---

## Design choices

**A tenant CR in the org workspace, reconciled straight to Keycloak.** The obvious
alternative is to keep IDPC as the single source of truth and have the operator merge
tenant intent into it. That adds a hop, copies the client secret into `root:orgs` —
creating a confused-deputy path on `clientSecretRef` — and leaves two conceptual writers
on one field. None of it buys anything, because the operator already holds Keycloak admin
rights and can reconcile from the org CR directly. IDPC keeps platform-managed clients,
realm flags, and the local Dex seed.

**A purpose-built type, not a reuse of `UpstreamIdentityProvider`.** Copying that struct
would copy the trust boundary problem with it. `OIDCUpstreamConfig` carried ~20 fields —
`validateSignatures`, `useJwksUrl`, `validatingPublicKey`, `clientAuthentication`, the
client-assertion set — that must never be tenant-settable. `IdPRegistration` exposes an
allowlist instead, and the operator builds the Keycloak representation field by field,
pinning the security-relevant ones to operator constants.

Once IDPC stopped managing upstream providers, those fields had no reader left anywhere —
setting them in the seed config was a silent no-op — so this PR deletes them together with
the mapper that used to consume them. What remains of `UpstreamIdentityProvider` /
`OIDCUpstreamConfig` is internal-only, appears in no CRD, and covers just the local Dex
seed and email-domain org linking. A reviewer should not have to work out which of those
knobs were still live.

**The operator is the only Keycloak Admin API client.** Every broker it writes carries
`platformMeshManagedBy=idpregistration`, so provenance is unambiguous: this broker came
from this CR. That is what makes the finalizer, the adoption guard, and per-org delete
tractable. The accepted cost is two reconcile paths for realm state — IDPC for platform
bits, `IdPRegistration` for tenant brokers — which is only tolerable because alias
ownership is explicit, hence the provenance marker and the reserved seed aliases.

Also worth flagging:

- **The FGA rules are not in this repo.** `data/coreModule.fga` is deliberately untouched;
  it is not loaded at runtime. The deployed module is `charts/security-operator/values.yaml`
  in helm-charts, so the owner-only `idpregistrations` rules live there rather than being
  duplicated here.
- **Clearing `emailDomainRouting.domains` deletes the linked Keycloak organization.**
  Destructive, documented on the API type, and not yet reference-counted (see limitations).
- **Domain ownership is not verified.** An org owner can claim any domain string. Routing
  is scoped to their own realm, so this is not an escalation, but it deserves a product
  decision before GA.
- **New flags**, all defaulted: `--idp-registration-secret-namespace`,
  `--keycloak-path-prefix`, `--idp-realm-deny-list`, `--idp-seed-config-file`.

---

## Why this is safe

The obvious objection is *"letting an org bring its own IdP lets someone become platform
admin."* It does not, and the reasons are structural rather than validations we could get
wrong:

| Attempt | Why it fails |
| --- | --- |
| Upstream asserts `groups: system:masters` | Keycloak validates the upstream token and issues **its own**; kcp never sees the upstream token. The WAC prefixes groups with `oidc:`, so `oidc:system:masters` matches nothing. |
| Token used against `root:orgs` or another org | Each `WorkspaceAuthenticationConfiguration` trusts exactly one issuer and is attached only to that org's WorkspaceTypes. |
| Read another org's client secret | The secret ref is **name-only** and resolves in the CR's own logical cluster. No tenant string ever becomes a lookup key in `root:orgs`. |
| Turn off signature validation | `validateSignatures=true` / `useJwksUrl=true` are operator-set constants with no corresponding API field, so there is nothing for a tenant to set. A hostile discovery document cannot change them either. |
| Become platform admin | Platform admin is a client certificate (`kcp-admin`), not an OIDC identity. There is no token that produces it. |
| Non-owner writes the CR | FGA owner-only, enforced at kcp via rebac-authz-webhook — not in the GraphQL gateway. |

Full threat model in the design doc, section *"Why this is safe."*

---

## Suggested review order

The diff is large, but most of it is generated schemas and Keycloak client plumbing.
The parts that deserve real attention, in order:

1. **[`apis/core/v1alpha1/idpregistration_types.go`](apis/core/v1alpha1/idpregistration_types.go)** — the tenant-facing API and its field
   allowlist. This is the trust boundary; everything else enforces it.
2. **`internal/webhook/idpregistration_validation_webhook.go`** — admission rules:
   HTTPS-only discovery, alias immutability, `spec.alias == metadata.name`, reserved seed
   aliases, manual OIDC requires `jwksUrl`.
3. **`internal/webhook/idpregistration_client_secret.go`** — the write-only `clientSecret`
   is moved into a Secret. Rejects adopting a secret that is unlabeled or owned by a
   different registration (secret-hijack guard).
4. **`internal/util/httpurl.go`** — shared SSRF guard: scheme allowlist, RFC1918 /
   loopback / link-local blocklist, DNS resolution check, redirect policy. Applied at
   admission **and** at fetch time in `pkg/clientreg/keycloak/discovery.go`, including
   re-validation of the endpoints returned by the discovery document.
5. **`internal/subroutine/idpregistration/subroutine.go`** — the reconcile loop. Worth
   checking: the provenance guard before adopting an existing alias, fail-closed finalize,
   and deny-listed realms terminating via `subroutines.Stop()` rather than retrying forever.
6. **`pkg/clientreg/keycloak/`** — the Admin API client: broker CRUD, discovery fetch,
   Organizations for email-domain routing, and mapper sync (desired set is empty, extras
   deleted every reconcile).

Generated / mechanical, safe to skim: `zz_generated.deepcopy.go`, `config/crd/**`,
`config/resources/**`, `test/setup/**`, `go.mod` / `go.sum` churn.

---

## Status API contract

Status is condition-based only — no bespoke `ready` / `message` fields that could drift
from the lifecycle-managed conditions. Portal UI builds on this, so it is worth agreeing
on now:

| Consumer need | Field |
| --- | --- |
| Ready indicator | `status.conditions[?(@.type=='Ready')]` |
| Keycloak / operator error text | `status.conditions[?(@.type=='IdPRegistration')].message` |
| Redirect URI to configure upstream | `status.redirectUri` |
| Linked org / domains | `status.organizationId`, `status.linkedEmailDomains` |

See [`docs/design/upstream-idp-ui-integration-plan.md`](docs/design/upstream-idp-ui-integration-plan.md).

---

## Known limitations (deliberate follow-ups, not oversights)

| Gap | Rationale for deferring |
| --- | --- |
| No reference-count guard before deleting a shared Keycloak organization | Needs a product decision: block, reference-count, or document as destructive. Only reachable when two registrations share routing semantics. |
| Collision with a pre-existing Keycloak broker surfaces at reconcile, not admission | Cross-org collisions cannot happen — each org has its own realm. Within a realm, admission would need a Keycloak Admin API call per request. Reserved seed aliases *are* blocked at admission, and for anything else the provenance guard refuses to adopt a broker it does not own and reports it on `status.conditions`. |
| No integration test proving FGA denies a non-owner | Needs envtest plus rebac-authz-webhook; the model itself ships and is exercised in helm-charts. |
| `internal/subroutine/idpregistration` unit coverage is thin | Deep reconcile tests need a sizable Keycloak mock. Security-critical paths (SSRF, secret ownership, provenance, config) do have unit tests. |

---

## Cross-repo

Requires **[helm-charts #2485](https://github.com/platform-mesh/helm-charts/pull/2485)**
for the `IdPRegistration` APIResourceSchema and the FGA `coreModule` rules. Merge or
release that alongside this PR — the FGA type is fail-closed, so a missing type looks
exactly like a permissions bug. (Supersedes helm-charts #2282, now closed.)

Mind the release ordering: #2485 ships `security-operator-crds` 0.6.2 with the schema, but
`charts/security-operator` still pins the published 0.6.0, which predates `IdPRegistration`.
That pin has to be bumped once 0.6.2 is released — until then the schema never reaches a
cluster that installs the parent chart.

Portal follow-ups (separate PRs, not blocking):
[portal-ui-lib #606](https://github.com/platform-mesh/portal-ui-lib/pull/606) →
platform-mesh-operator (ContentConfiguration) → portal (dep bump).

---

## Test plan

- [x] `task test-security-operator`
- [x] `task lint-security-operator`
- [x] Unit tests for SSRF helpers, discovery validation, secret-ownership rejection,
      reserved seed aliases, broker provenance guards
- [ ] `task cover` for `internal/subroutine/idpregistration` (below the 80% threshold; see limitations)
- [ ] Manual on kind local-setup: create an `IdPRegistration`, verify the Keycloak broker
      appears and `status.redirectUri` is populated
- [ ] Manual: non-owner create is denied by FGA


philtk79:
Closing so we can recut this as a stacked draft chain from the fork. The work is not dropped — it will be split into smaller reviewable PRs.

PR #354, opened 2026-09-08
feat(search): filter accounts by FGA relation

### LINKED ISSUE #372: story: Add `filter.fga_role` Query Parameter for FGA-Based Pre-filtering
### User Story

As a portal user,
I want to filter search results by my FGA role (e.g. filter.fga_role=owner)
so that I only see resources belonging to accounts I own, without having to manually identify which accounts those are.

## Context

Currently there is no way to say "only show me resources I own" in a single request. Users who are owners of a subset of accounts see every resource in OpenSearch, with post-hoc FGA filtering silently trimming results. This limits queries for the users role dependencies.

## Solution Path

Add `filter.fga_role=<relation>` as a new recognized query parameter distinct from the generic `filter.<field>` family. When present, it is resolved at query time and before the OpenSearch request is issued. We can do this by fetching the set of FGA objects for which the user holds the requested relation, then injecting those as an OpenSearch `terms` filter on `filterable_fields.account_fga_object` (or a similar field mapping).

Example:

```
GET /rest/v1/search?q=my-app&filter.fga_role=owner
```

Returns only resources belonging to accounts where the authenticated user has the `owner` relation in OpenFGA.

## Out of Scope

- Changing the post-query FGA `BatchCheck` behavior.
- Exposing `filter.fga_role` on the `/resources` endpoint.
- Multi-value `filter.fga_role` (single relation only for this ticket).

---

## Open Question

Should `filter.fga_role` be validated against a known set of relations (e.g. `owner`, `viewer`, `editor`) or left open to any string?

### Acceptance Criteria

- [ ] `GET /rest/v1/search?filter.fga_role=owner` returns only resources where the authenticated user holds the `owner` relation in OpenFGA for the associated account.
- [ ] `GET /rest/v1/search?filter.fga_role=viewer` likewise scopes to viewer-accessible resources.
- [ ] Combining `filter.fga_role` with other `filter.*` params works correctly (AND semantics).
- [ ] If the user holds the requested role on zero accounts, the response is an empty result set (no scan, fast return).
- [ ] `filter.fga_role` with an unrecognized/invalid relation value returns HTTP 400.
- [ ] The existing post-query FGA `BatchCheck` still runs as a safety net — `filter.fga_role` is an optimization, not a security bypass.
- [ ] Unit tests cover: role resolved to a non-empty list, role resolved to empty list, invalid relation, combined with other filters.
- [ ] No change to existing behavior when `filter.fga_role` is absent.

### Parent Epic

_No response_

### Additional Context

_No response_

hellsontime:
- Add `filter.fga_role` to restrict the account pre-filter using any relation from the active OpenFGA model.
- Require an explicit `resource` or `resources` target when using `filter.fga_role`.
- Return HTTP 400 for malformed or schema-unknown relations and bind the relation to pagination cursors.
- Preserve the final per-result FGA authorization check.
- Fix the pre-existing SearchIndex resource default: `searchindexes` → `searchindices`.
- Fix `filterable_fields.account_fga_object` for Account resources: an Account is now filtered by itself instead of by its parent Account (platform-mesh/backlog#372 review comment). Note that `account_name` / `account_id` are deliberately left as-is — they are display fields naming the Account a resource lives in, not authorization keys.

Closes platform-mesh/backlog#372

## Testing

- `go test -count=1 ./...`
- `go build ./...`
- `task lint` (`0 issues`)
- Manual E2E on `kind-platform-mesh` covering default, `owner`, and `member` relations; document filters; cursor isolation; unknown relations; account pre-filtering; and final per-result authorization.

### DIFF
--- a/services/search-service/internal/router/router.go
+++ b/services/search-service/internal/router/router.go
@@ -22,6 +22,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"regexp"
 	"strconv"
 	"strings"
 
@@ -31,6 +32,8 @@ import (
 	appcontext "go.platform-mesh.io/search-service/internal/context"
 	"go.platform-mesh.io/search-service/internal/service/search"
 )
+
+var fgaRelationPattern = regexp.MustCompile(`^[^:#@\s]{1,50}$`)
 
 type SearchService interface {
 	Search(ctx context.Context, req search.SearchRequest) (search.SearchResponse, error)
@@ -272,6 +275,9 @@ func parseSearchFilters(values map[string][]string) (map[string][]string, string
 	fgaRole := strings.TrimSpace(entries[0])
 	if fgaRole == "" {
 		return nil, "", fmt.Errorf("filter.fga_role must not be empty")
+	}
+	if !fgaRelationPattern.MatchString(fgaRole) {
+		return nil, "", fmt.Errorf("filter.fga_role has an invalid format")
 	}
 
 	delete(filters, "fga_role")
--- a/services/search-service/internal/router/router.go
+++ b/services/search-service/internal/router/router.go
@@ -69,7 +69,7 @@ func CreateRouter(svc SearchService, mws []func(http.Handler) http.Handler) *chi
 
 		resource := strings.TrimSpace(r.URL.Query().Get("resource"))
 
-		filters, err := parseFilters(r.URL.Query())
+		filters, fgaRole, err := parseSearchFilters(r.URL.Query())
 		if err != nil {
 			http.Error(w, err.Error(), http.StatusBadRequest)
 			return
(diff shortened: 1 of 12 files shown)


services/search-service/internal/router/router.go
```diff
@@ -32,6 +33,8 @@ import (
 	"go.platform-mesh.io/search-service/internal/service/search"
 )
 
+var fgaRelationPattern = regexp.MustCompile(`^[^:#@\s]{1,50}$`)
```
pteich, line 37: Is it really a good idea to have a deny list here (not :#@\s) instead of an allow list like `[a-zA-Z0-9_][a-zA-Z0-9_-]*`?

pteich (dismissed):
just one nitpick - lgtm in general

FWuermse (dismissed):

services/search-service/internal/router/router.go
hellsontime: The pattern is not something I came up with, it is a verbatim copy of OpenFGA's own validation for ListObjectsRequest.relation (`^[^:#@\s]{1,50}$` in the generated proto validators). Florian confirmed that account models can use arbitrary relation names, so the OpenFGA schema is the source of truth here. The only purpose of the local check is to return a clean 400 for values OpenFGA would reject anyway, instead of surfacing its InvalidArgument as a 500.

A stricter allow list would reject relations that OpenFGA itself accepts. The value is also never interpolated into a query: it goes into a typed gRPC field and into the cursor hash, so the deny list does not open an injection path. I would keep it aligned with OpenFGA, but happy to tighten it if we decide to restrict relation naming on the platform side.

aaronschweig (approved):

PR #357, opened 2026-09-10
feat(ui-api): expose category, labels, mainLink and serviceLevel

### LINKED ISSUE #352: feat: Add missing ProviderMetadata fields to expose full catalog card data
### Context

The `ProviderMetadata` CRD is the data source for provider cards in the marketplace UI. The marketplace UI already has TypeScript models and HTML templates wired up for several fields that are **not yet present in the Go type**. Because the CRD schema prunes unknown fields, these values are stripped on admission and never reach the UI.

The fields are already modeled in the UI as `// not supported yet` in [`src/app/models/provider-metadata.ts` (L134–142)](https://github.com/platform-mesh/marketplace-ui/blob/main/src/app/models/provider-metadata.ts#L134-L142).

> **Note:** `verification` is tracked separately in its own issue — the struct design requires a dedicated discussion. See the epic for the linked issue.

### Missing fields

| Field | UI location | Template reference |
|---|---|---|
| `category` | Subtitle below provider name | [`provider-detail-dialog.component.html` L9](https://github.com/platform-mesh/marketplace-ui/blob/main/src/app/components/provider/provider-detail-dialog/provider-detail-dialog.component.html#L9) |
| `mainLink` | Primary "Open" action button (top-right) | [`provider-detail-dialog.component.html` L19–33](https://github.com/platform-mesh/marketplace-ui/blob/main/src/app/components/provider/provider-detail-dialog/provider-detail-dialog.component.html#L19-L33) |
| `labels` | Colored info-label chips in the card header | [`provider-detail-dialog.component.html` L91–100](https://github.com/platform-mesh/marketplace-ui/blob/main/src/app/components/provider/provider-detail-dialog/provider-detail-dialog.component.html#L91-L100) |
| `serviceLevel` | "Service Level:" row under "How to get help?" | [`provider-detail-dialog.component.html` L181](https://github.com/platform-mesh/marketplace-ui/blob/main/src/app/components/provider/provider-detail-dialog/provider-detail-dialog.component.html#L181) |

### Work

1. Add the 4 fields + required helper types (`Label`, `ServiceLevel`) to [`apis/ui/v1alpha1/providermetadata_types.go`](https://github.com/platform-mesh/platform-mesh/blob/main/apis/ui/v1alpha1/providermetadata_types.go)
2. Regenerate the CRD manifest (`make generate manifests` or equivalent)
3. Update a provider on a local cluster with values for all 4 fields and verify each renders correctly in the marketplace UI

### Acceptance Criteria

- [ ] `ProviderMetadata` spec accepts `category`, `labels`, `mainLink`, and `serviceLevel`
- [ ] CRD manifest is regenerated and committed
- [ ] Patching a provider with all 4 fields stores them (no pruning warnings)
- [ ] Each field renders in the marketplace detail card as the UI already expects

gkrajniak:
refactor(ui-api): simplify ProviderMetadataSpec surface
## What changed
- Removed the `Verification` type and `ProviderMetadataSpec.Verification`
  field (reverted; superseded by the schemaless `data` blob).
- Added `Main bool` to the shared `Link` type and removed
  `ProviderMetadataSpec.MainLink`; a link is now marked as the main one via
  `links[].main` instead of a separate field.
- Removed the `Label` type and `ProviderMetadataSpec.Labels` field; provider
  chips are now derived entirely from `Tags []string`.
- Removed `ProviderMetadataSpec.Category`.
- Removed `ProviderMetadataSpec.Type`.

## Why
Several of these typed fields either duplicated data that already lives in
the schemaless `Data *apiextensionsv1.JSON` blob (category, provider,
verification) or were superseded by a simpler representation on an existing
field (a main link folded into `links`, labels derived from `tags`).
Consolidating them keeps the typed CRD schema minimal and pushes UI-only /
still-evolving metadata into the free-form `data` field, matching how
`verification` was already modeled before this change.

## Details
- Regenerated deepcopy (`zz_generated.deepcopy.go`), CRDs, and the
  extension-manager-operator's APIExport / APIResourceSchema manifests via
  `task generate` after each field removal.
- Verified: `apis` and `operators/extension-manager-operator` build and their
  test suites pass after every change in this set.


aaronschweig:
Since this information is rather administrator individual I was thinking of using the `.spec.data` property for specific fields rather than overpopulating the `ProviderMetadata` resource directly. I think the default resource should maybe not contain every item possible but only something that we say is the bare minimum and then via the marketplace-ui extensions specific fields from the data field can displayed however people like.

apis/ui/v1alpha1/providermetadata_types.go
```diff
@@ -53,12 +53,53 @@ type DetailViewExtension struct {
...
+	Label string `json:"label"`
+	// Status is the badge color, mapped to a Fundamental ObjectStatus.
+	// +kubebuilder:validation:Enum=positive;critical;negative;informative;neutral
+	Status string `json:"status,omitempty"`
+	// Icon is an optional SAP-icons glyph name; the UI defaults to a verified
+	// glyph when it is absent.
+	Icon string `json:"icon,omitempty"`
+	// Hint is an optional tooltip shown on the badge.
+	Hint string `json:"hint,omitempty"`
+}
```
aaronschweig, line 93: That is very provider specific IMHO, PM does not have a notion of a "verified" provider

apis/ui/v1alpha1/providermetadata_types.go
```diff
@@ -53,12 +53,53 @@ type DetailViewExtension struct {
...
+
 // ProviderMetadataSpec defines the desired state of ProviderMetadata.
 type ProviderMetadataSpec struct {
 	Tags []string `json:"tags,omitempty"`
 
 	DisplayName string `json:"displayName"`
 	Description string `json:"description,omitempty"`
+	Type        string `json:"type,omitempty"`
+	Category    string `json:"category,omitempty"`
```
aaronschweig, line 102: what is the difference between a Catogory and the already present field `Tags`?

apis/ui/v1alpha1/providermetadata_types.go
```diff
@@ -67,9 +108,13 @@ type ProviderMetadataSpec struct {
 	Icon          *Icon                 `json:"icon,omitempty"`
 
 	Links                    []Link                `json:"links,omitempty"`
+	MainLink                 *Link                 `json:"mainLink,omitempty"`
```
aaronschweig, line 111: what is the difference between main link and links? Why is there a need to differentiate

apis/ui/v1alpha1/providermetadata_types.go
```diff
@@ -67,9 +108,13 @@ type ProviderMetadataSpec struct {
...
 	PreferredSupportChannels []Link                `json:"preferredSupportChannels,omitempty"`
 	HelpCenterData           []Link                `json:"helpCenterData,omitempty"`
 	DetailViewExtensions     []DetailViewExtension `json:"detailViewExtensions,omitempty"`
+	Labels                   []Label               `json:"labels,omitempty"`
```
aaronschweig, line 115: Labels, Tags, Categories, sounds all very similar to me - any chance to consolidate?

apis/ui/v1alpha1/providermetadata_types.go
gkrajniak: @lpgarzonr

apis/ui/v1alpha1/providermetadata_types.go
gkrajniak: **Main link** is a link used in the provider detail dialog

image

but true, it we could introduce a flag main in the Links, and the n the thing goes together 


type Link struct {
	DisplayName string `json:"displayName,omitempty"`
	URL         string `json:"url,omitempty"`
	**Main        bool `json:"url,omitempty"`**
}

@lpgarzonr

apis/ui/v1alpha1/providermetadata_types.go
gkrajniak: the samwe thing with the Labels we could have some diferentiator for type/category/

Screenshot 2026-09-11 at 10 03 22

the tags we don't use in the ui, could be cleaned up

@lpgarzonr ?

apis/ui/v1alpha1/providermetadata_types.go
gkrajniak: tags we don't use

apis/ui/v1alpha1/providermetadata_types.go
lpgarzonr: The separation is intentional `mainLink` maps to a single prominent "Open" CTA button in the detail view header, which is semantically different from a list of resource links in the body. Keeping it as a dedicated field makes the "**exactly one**" invariant explicit in the schema.

That said, I'm also fine folding it into `links` with a `main: bool` flag if the team prefers a simpler model. The trade-off is that the constraint becomes implicit  nothing in the schema prevents two entries with `main: true`, so we'd need runtime filtering in the UI and a validation webhook later to enforce it. With the dedicated field, the invariant is self documenting and the webhook would only be needed to reject specs that set both.

Worth noting: `links` is currently not queried in the GraphQL query and not rendered anywhere in the marketplace-ui. Before deciding on the model, we should clarify its intended use if it's meant to show a "Resources" section in the detail view, that affects which approach makes more sense.

apis/ui/v1alpha1/providermetadata_types.go
lpgarzonr: you're right, PM has no notion of a verified provider. We can move it to spec.data. then our Portal can store {"badge": {"text": "HyperText", "status": "informative"}} there, and the marketplace-ui can read it generically. No Go type change needed.
We can though also consider to call it smt more generic and show it always under the title but not Verification.

apis/ui/v1alpha1/providermetadata_types.go
lpgarzonr: They serve distinct purposes in the UI and aren't interchangeable.

- `category` is a single string — renders as the page subtitle in the detail dialog, as a "Category: …" footer on the catalog card, and drives the category dropdown filter.
- `tags` is currently not queried in the GraphQL query and not rendered anywhere in the marketplace-ui. However, it is possible that there are already consumers using it as plain strings, so changing the type would be a breaking change.

Recommendation: **drop `Labels []Label` from this PR**. In the old portal Extension model, `Label` is just `{Title string}` — structurally identical to `tags []string`. The marketplace-ui's `buildLabels` already auto-derives chip color from the title via `mapToColorCategory`, so updating it to read `spec.tags` would render colored info-label chips with no schema change and no breaking change to existing consumers.

lpgarzonr: What do you think of making the filters fully configurable via the ContentConfiguration `context`? The idea is that consumers who need filtering define their own filter dimensions there, pointing to values in `spec.data`:
```json
"uiConfig": {
  "filters": [
    { "label": "Category", "dataPath": "spec.data.category" },
    { "label": "Provider",  "dataPath": "spec.data.provider"  }
  ]
}
```
Providers that need to be filterable would then set the corresponding values in `spec.data`:
```yaml
spec:
  data:
    category: "Cloud native development" // or even smt more reach object based on what we need to show 
    provider: "community"
```
This way consumers that don't need filters simply don't configure them, and the `ProviderMetadata` API stays clean with no hardcoded `category` or `type` fields needed. The tradeoff vs typed fields is that `spec.data` is free-form so there's no schema validation, but I think the flexibility outweighs that for this use case. I'll open a separate issue to implement the configurable filters and as part of it we can remove `category` and `type` from this PR.

apis/ui/v1alpha1/providermetadata_types.go
gkrajniak: done

gkrajniak: done

apis/ui/v1alpha1/providermetadata_types.go
gkrajniak: done

aaronschweig: actually there are kubebuilder validation rules, where we can verify such things. So this could still be implemented with a simple validation rule

apis/ui/v1alpha1/providermetadata_types.go
gkrajniak: just a small adjustment to indicate well the data path =>

```
            "uiConfig": {
              "filters": [
                {
                  "label": "Category",
                  "providerMetadataPath": "spec.data.category"
                },
                {
                  "label": "Provider",
                  "providerMetadataPath": "spec.data.provider"
                }
              ]
            }
```

apis/ui/v1alpha1/providermetadata_types.go
```diff
@@ -70,6 +84,7 @@ type ProviderMetadataSpec struct {
 	PreferredSupportChannels []Link                `json:"preferredSupportChannels,omitempty"`
 	HelpCenterData           []Link                `json:"helpCenterData,omitempty"`
 	DetailViewExtensions     []DetailViewExtension `json:"detailViewExtensions,omitempty"`
+	ServiceLevel             ServiceLevel          `json:"serviceLevel,omitempty"`
```
gkrajniak: here would be a question as well about the ServiceLevel is it something general, hsp specific ... etc.?
@lpgarzonr @aaronschweig

lpgarzonr: I think it make sense for any provider, but perhaps make the model more general like an string

apis/ui/v1alpha1/providermetadata_types.go
```diff
@@ -38,6 +38,7 @@ type Image struct {
 type Link struct {
 	DisplayName string `json:"displayName,omitempty"`
 	URL         string `json:"url,omitempty"`
+	Main        bool   `json:"main,omitempty"`
```
aaronschweig, line 41:
```suggestion
// Links are provider-owned links. At most one link may be marked as the
// main link via `main: true`.
// +kubebuilder:validation:XValidation:rule="self.filter(l, has(l.main) && l.main).size() <= 1",message="only one link may be marked as main"
type Link struct {
DisplayName string `json:"displayName,omitempty"`
URL         string `json:"url,omitempty"`
	Main        bool   `json:"main,omitempty"`
```

I think we should include that validation rule to mitigate the risk @lpgarzonr mentioned

gkrajniak: done

apis/ui/v1alpha1/providermetadata_types.go
```diff
@@ -66,10 +67,16 @@ type ProviderMetadataSpec struct {
...
+	// Links are provider-owned links. At most one link may be marked as the
+	// main link via `main: true`.
+	// +kubebuilder:validation:MaxItems=20
+	// +kubebuilder:validation:XValidation:rule="self.filter(l, has(l.main) && l.main).size() <= 1",message="only one link may be marked as main"
+	Links []Link `json:"links,omitempty"`
+
 	PreferredSupportChannels []Link                `json:"preferredSupportChannels,omitempty"`
 	HelpCenterData           []Link                `json:"helpCenterData,omitempty"`
 	DetailViewExtensions     []DetailViewExtension `json:"detailViewExtensions,omitempty"`
+	ServiceLevel             string                `json:"serviceLevel,omitempty"`
```
gkrajniak, line 79: here?

lpgarzonr (approved):

lpgarzonr (approved):

PR #358, opened 2026-09-11
feat: IdPRegistration for org upstream OIDC IdPs

philtk79:
## Why

Org owners can't configure an upstream OIDC IdP today — the only place upstream providers can live is `IdentityProviderConfiguration` (IDPC) in `root:orgs`, a workspace tenants must not write to. This PR adds a tenant-owned `IdPRegistration` CR end to end: the API, PMO install path, admission, and security-operator reconciliation to a Keycloak OIDC broker.

KcpSetup in platform-mesh-operator is what actually installs `core.platform-mesh.io` into org workspaces. Account- and security-operator `config/resources` copies are generated schema output for envtest/apigen (same pattern as `invites`), not a second install path.

## Summary

- `IdPRegistration` CRD types + scheme registration
- Generated CRDs / APIResourceSchemas in account-operator and security-operator (`v260907-08269bea`)
- platform-mesh-operator KcpSetup: `apiresourceschema-idpregistrations` + `idpregistrations` on `apiexport-core.platform-mesh.io`
- Validating/mutating admission webhooks + PMO KCP webhook configs and CA wiring
- Reconcile controller/subroutine: Keycloak OIDC broker CRUD, mapper sync, organization linkage
- `oidc:` groups-claim prefix on org WACs
- `docs/design/upstream-identity-provider-registration.md` — design rationale
- `internal/util/httpurl.go`: SSRF guard (scheme allowlist, RFC1918/loopback block, DNS check, redirect policy)
- `pkg/clientreg/keycloak` additions: broker CRUD, discovery fetch, Organizations API, mapper sync

Does **not** include seed or Portal ContentConfiguration.

## Authorization

Owner-only OpenFGA relations for `idpregistrations` live in helm `coreModule` — see [helm-charts#2495](https://github.com/platform-mesh/helm-charts/pull/2495). Until that lands, auth **fails closed**: org owners cannot write the CR; `kcp-admin` still can.

## Follow-ups

1st of 2 platform-mesh PRs. Merge this before seed.

- #361 seed path (local/dev bootstrap through the same CR shape)
- Portal ContentConfiguration (separate PR)
- [helm-charts#2495](https://github.com/platform-mesh/helm-charts/pull/2495) chart webhook registration + FGA `coreModule`

Former layer PRs #359 and #360 are folded into this one.

## Ticket

[platform-mesh/backlog#286](https://github.com/platform-mesh/backlog/issues/286)

## Test plan

[...]

PR #359, opened 2026-09-11
feat: admission webhooks for IdPRegistration

philtk79:
## Why

Types alone don't protect anything — without admission, a tenant could write an `IdPRegistration` with an SSRF-able discovery URL, a non-HTTPS endpoint, or write access it shouldn't have, straight into the org workspace. This PR wires up validating and mutating webhooks so the guard rails from #358 actually run on every write, and prefixes WAC groups with `oidc:` so an upstream `groups` claim cannot match platform identities.

Helm-charts registers the in-cluster webhook Service. Org-workspace writes go through KCP, so platform-mesh-operator KcpSetup must also apply the KCP `ValidatingWebhookConfiguration` / `MutatingWebhookConfiguration` and inject the security-operator CA. Those manifests and the CA wiring are ported from the archived `platform-mesh-operator` `feat/upstream-idp` branch. They share `feature-disable-idp-webhook` with the existing IDPC webhook.

## Summary

- Validating webhook: required fields, discovery-vs-manual mutual exclusion, HTTPS-only, SSRF block via #358's guard, email-domain format
- Mutating webhook: strips `clientSecret` into a Secret before persist
- `workspace_authorization.go`: prefix WAC groups with `oidc:`
- Registered in `cmd/operator.go`
- platform-mesh-operator: KCP validating + mutating webhook manifests for `idpregistrations`, CA copied from the security-operator secret
- No reserved-alias check — the seed path creates CRs with the same aliases that check would reject, and seed writes aren't conceptually subject to admission

## Follow-ups

2nd of 4. Merge after #358. This Files changed tab includes #358; layer-only diff: https://github.com/philtk79/platform-mesh/compare/feat/idp-1-foundation...feat/idp-2-webhooks

Pairs with [helm-charts#2495](https://github.com/platform-mesh/helm-charts/pull/2495) (in-cluster Service). KCP admission configs are in this PR.

## Ticket

[platform-mesh/backlog#286](https://github.com/platform-mesh/backlog/issues/286)

## Test plan

- [x] `go test ./operators/security-operator/internal/webhook/`
- [x] `go test ./operators/platform-mesh-operator/pkg/subroutines/`
- [ ] Manual: apply an `IdPRegistration` with an HTTP discovery URL, confirm rejection


philtk79:
Superseded by #358 — webhooks and PMO KCP webhook wiring are now part of the complete IdPRegistration feature PR. Closing this layer split.

PR #366, opened 2026-09-15
feat(search-operator): add searchable-resources flag

FWuermse:
This commit restores some code that got lost during a rebase as part of the cleanup PR (#346).

- Provider per resource must no longer be configured
- Flats for GVKs to be reconciled was never populated
- `kubeconfig-copy.sh` was outdated and not working for podman


operators/search-operator/cmd/operator.go
```diff
@@ -121,9 +121,9 @@ func RunController(_ *cobra.Command, _ []string) { // coverage-ignore
 		log.Fatal().Err(err).Str("controller", "SearchIndex").Msg("unable to create controller")
 	}
 
-	for _, gvk := range operatorCfg.SearchableResources {
+	for _, GVK := range operatorCfg.SearchableResources {
```
aaronschweig: having that uppercase is not really idiomatic go, lowercase would be better

FWuermse: Hi, I usually refer to the casing recommended in go wiki because those are the defaults for strict linting rules: https://go.dev/wiki/CodeReviewComments#initialisms

operators/search-operator/internal/config/config.go
```diff
@@ -25,28 +25,43 @@ import (
...
+	target *[]schema.GroupVersionKind
 }
 
-func (g *GroupVersionKindProvider) Unmarshal(s string) error {
-	s = strings.Trim(s, "{}")
-	parts := strings.Split(s, ",")
-	if len(parts) != 4 {
-		return fmt.Errorf("expected 'group,version,kind,provider', got %q", s)
+func (g *gvkSliceValue) String() string {
+	parts := make([]string, len(*g.target))
```
pteich: i feel very uncomfortable with this amount of dereferencing, why `g.target` needs to be a pointer?

operators/search-operator/cmd/operator.go
aaronschweig: with this I also agree, I just don't think this is a case of an initialism. E.g. URL is a well known and accepted abbreviation, GVK not really.

operators/search-operator/internal/config/config.go
```diff
@@ -25,28 +25,43 @@ import (
...
+	for _, s := range strings.Split(val, ",") {
+		s = strings.TrimSpace(s)
+		if s == "" {
+			continue
+		}
+		parts := strings.SplitN(s, "/", 3)
+		if len(parts) != 3 {
+			return fmt.Errorf("invalid GVK %q: expected group/version/kind", s)
+		}
+		*g.target = append(*g.target, schema.GroupVersionKind{
```
pteich: Again so much de-referencing of the pointer. Can we avoid this?

pteich (approved):

akafazov (approved):

PR #371, opened 2026-09-17
Fix search-service fail in some cases

pteich:
The current version of the search-services failed with a status 500 on several search requests. One example:
```
curl --request get \
  --url 'https://search.portal.localhost:8443/rest/v1/search?q=test&resource=components&page=1&limit=10' \
  --header 'accept: application/json, text/plain, */*' \
  --header 'authorization: Bearer <my-keycloak-token>' 
  ```

The reason is that some indexed documents contain numerical and text values and the search query can't handle this.
The main fix is to add a 
```
"lenient": true,
```
to the query to ignore fields that have a different type not matching the query.

In addition, the PR adds some better error reports for similar future issues and a small refactoring to re-use a string sorting function.

### CODE REVIEW

File: services/search-service/internal/clients/opensearch/client.go [hunk 1]
```diff
@@ -312,35 +336,35 @@ func (c *Client) Search(ctx context.Context, query search.OpenSearchQuery) (sear
 		slices.Sort(values)
 	}
 
+	// Process OpenSearch shard errors: if some shards fail, hits and TotalCount
+	// can be partial. If all shards fail (e.g. mapping errors) we need to dedupe the messages
+	var shardFailures []string
+	seenFailures := make(map[string]struct{}, len(payload.Shards.Failures))
```
FWuermse: Let's use "k8s.io/apimachinery/pkg/util/sets" to be consistent.

services/search-service/internal/strings/strings.go
```diff
@@ -0,0 +1,48 @@
...
+
+// DedupeSorted trims every value, drops the empty ones, removes duplicates and
+// sorts what remains. The result is never nil.
+func DedupeSorted(values []string) []string {
+	out := make([]string, 0, len(values))
+	if len(values) == 0 {
+		return out
+	}
+
+	seen := make(map[string]struct{}, len(values))
```
FWuermse: Also here we can use "k8s.io/apimachinery/pkg/util/sets".

services/search-service/internal/clients/opensearch/client.go
pteich: good catch! fixed

services/search-service/internal/strings/strings.go
pteich: fixed

FWuermse (dismissed):

FWuermse (approved):

PR #372, opened 2026-09-17
fix(ci): change runner to ubuntu-latest-large and configure local domain resolution

akafazov:
Supports e2e tests for the `platform-mesh-operator` in subsequent PR.


jakobmoellerdev:
please consider the cost impact here on neonephos. it will cost the enterprise real money. can we optimize the setup instead?

akafazov:
closing this, as operator e2e tests require `ubuntu-latest-large` runner

